Issue #1, opened 2025-10-09
Implement x/fibre module


rootulp:
## Context

https://github.com/celestiaorg/fibre-da-spec/pull/3

## Steps

- [x] Implement protos for x/fibre (.proto files, generated .pb.go files): https://github.com/celestiaorg/celestia-app/pull/5794 (merged on celestia-app repo then copied to celestia-app-fibre repo via `feature/fibre` branch)
- [x] Implement stateless validation `ValidateBasic` for all the types (including unit tests): https://github.com/celestiaorg/celestia-app-fibre/pull/2
- [x] Implement keepers for state related changes (including unit tests): https://github.com/celestiaorg/celestia-app-fibre/pull/4
- [x] Implement QueryServer: https://github.com/celestiaorg/celestia-app-fibre/pull/32
- [x] Implement MsgServer + message handlers (including unit tests + events): https://github.com/celestiaorg/celestia-app-fibre/pull/34
- [x] Implement CLI commands: https://github.com/celestiaorg/celestia-app-fibre/pull/35
- [x] Implement `BeginBlock` automatic processing of withdrawals, automatic pruning of withdrawals past time window
    - https://github.com/celestiaorg/celestia-app-fibre/pull/41
    - https://github.com/celestiaorg/celestia-app-fibre/pull/43
- [x] https://github.com/celestiaorg/celestia-app-fibre/issues/47
- [ ] A fork of mempool state needs to be persisted so that ValidatePaymentPromises can handle multiple payment promises in one block. https://github.com/celestiaorg/celestia-app/issues/6824

rootulp:
IMO we can close this tracking issue and keep https://github.com/celestiaorg/celestia-app/issues/6824 open.

PR #2, opened 2025-10-09
feat(x/fibre): `ValidateBasic` methods

rootulp:
Part of https://github.com/celestiaorg/celestia-app-fibre/issues/1

To-do:
- [x] Verify that this actually implements all the stateless validation described in the spec
- [x] ~~Verify the "stateless signature" checks. Validator signatures must use ed25519 and we want to use the Go std lib implementation for that~~ the spec still has signature verification as a stateful thing so this isn't needed yet.

### DIFF
--- a/x/fibre/types/msgs.go
+++ b/x/fibre/types/msgs.go
@@ -1,0 +1,194 @@
+package types
+
+import (
+	"fmt"
+
+	errorsmod "cosmossdk.io/errors"
+	"github.com/celestiaorg/go-square/v2/share"
+	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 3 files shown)


x/fibre/types/msgs.go
```diff
@@ -0,0 +1,166 @@
...
+
+	if msg.SignerPublicKey == nil {
+		return errorsmod.Wrap(sdkerrors.ErrInvalidPubKey, "signer public key cannot be nil")
+	}
+
+	pubKey, ok := msg.SignerPublicKey.GetCachedValue().(cryptotypes.PubKey)
+	if !ok {
+		return errorsmod.Wrap(sdkerrors.ErrInvalidPubKey, "failed to get cached public key")
+	}
+
```
evan-forbes, line 100: interesting and nice makes sense

evan-forbes (approved):

PR #6, opened 2025-10-13
feat(x/fibre): add `SignatureSet` for validator signature collection

Wondertan:
Adds `SignatureSet` to allow fibre client to await the necessary signature
thresholds which are count and power from `validator.Set`.

Depends on https://github.com/celestiaorg/celestia-app-fibre/pull/5

### DIFF
--- a/x/fibre/validator/signature_set.go
+++ b/x/fibre/validator/signature_set.go
@@ -9,13 +9,6 @@ import (
 	core "github.com/cometbft/cometbft/types"
 )
 
-var (
-	// ErrNotEnoughSignatures indicates that not enough signatures were collected to meet the count threshold.
-	ErrNotEnoughSignatures = fmt.Errorf("not enough signatures")
-	// ErrNotEnoughVotingPower indicates that the collected signatures don't have sufficient voting power.
-	ErrNotEnoughVotingPower = fmt.Errorf("not enough voting power")
-)
-
 // SignatureSet collects and validates signatures from validators.
 // It is safe for concurrent use.
 type SignatureSet struct {
@@ -78,18 +71,42 @@ func (ss *SignatureSet) Done() <-chan struct{} {
 }
 
 // Signatures returns all collected signatures if thresholds are met.
-// Returns [ErrNotEnoughSignatures] if count threshold is not met.
-// Returns [ErrNotEnoughVotingPower] if voting power threshold is not met.
+// Returns [NotEnoughSignaturesError] if either count or voting power threshold is not met.
+// The error contains the partially collected signatures and threshold information.
 func (ss *SignatureSet) Signatures() ([][]byte, error) {
 	ss.mu.Lock()
 	defer ss.mu.Unlock()
 
-	if len(ss.signatures) < ss.minRequiredSignatures {
-		return nil, ErrNotEnoughSignatures
-	}
-	if ss.votingPower < ss.minRequiredVotingPower {
-		return nil, ErrNotEnoughVotingPower
+	countNotMet := len(ss.signatures) < ss.minRequiredSignatures
+	powerNotMet := ss.votingPower < ss.minRequiredVotingPower
+	if countNotMet || powerNotMet {
+		return nil, &NotEnoughSignaturesError{
+			Collected:      ss.signatures,
+			RequiredCount:  ss.minRequiredSignatures,
+			CollectedPower: ss.votingPower,
+			RequiredPower:  ss.minRequiredVotingPower,
+		}
 	}
 
 	return ss.signatures, nil
+}
+
+// NotEnoughSignaturesError indicates that signature collection did not meet the required thresholds.
+// It contains the partially collected signatures and threshold information.
+type NotEnoughSignaturesError struct {
+	Collected      [][]byte
+	RequiredCount  int
+	CollectedPower int64
+	RequiredPower  int64
+}
+
+func (e *NotEnoughSignaturesError) Error() string {
+	switch {
+	case len(e.Collected) < e.RequiredCount:
+		return fmt.Sprintf("not enough signatures: collected %d, required %d", len(e.Collected), e.RequiredCount)
+	case e.CollectedPower < e.RequiredPower:
+		return fmt.Sprintf("not enough voting power: collected %d, required %d", e.CollectedPower, e.RequiredPower)
+	default:
+		panic("unreachable")
+	}
 }
--- a/x/fibre/validator/signature_set.go
+++ b/x/fibre/validator/signature_set.go
@@ -1,7 +1,6 @@
 package validator
 
 import (
-	"context"
 	"crypto/ed25519"
 	"fmt"
 	"sync"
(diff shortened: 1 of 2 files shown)


evan-forbes:
can we point at the branch on #5 maybe? also we could turn on a merge queue here

x/fibre/validator/signature_set.go
```diff
@@ -0,0 +1,118 @@
...
+	if ss.validatorCount < ss.totalCount {
+		return
+	}
+
+	select {
+	case <-ss.done:
+	default:
+		close(ss.done)
+	}
+}
```
evan-forbes, line 92: could we infer this?

evan-forbes (dismissed):

x/fibre/validator/signature_set.go
Wondertan: Inferring requires a timeout, even in case where we know everyone responded and there is nothing left to wait on, unlike explicit miss handling.

evan-forbes: why pass a validator as an arg if its not used?

> Inferring requires a timeout

do we not have a timeout when the user waits for the signatures to come back?

Wondertan: Ah wait, you are right, there is an unused param there, which is from the older version. I will update it.

Wondertan: > do we not have a timeout when the user waits for the signatures to come back?

The timeout helps when validator doesn't respond, but if it responds with bad or no data, we need to mark the validator as responded without counting his voting power, so we don't have to wait for the timeout when we already know we can proceed. 

We don't have this validator response timeout defined anywhere actually. Implementation wise it is easy as the code is designed around context.

x/fibre/validator/signature_set.go
```diff
@@ -0,0 +1,117 @@
...
+		requiredBytesSigned: requiredBytesSigned,
+		requiredVotingPower: requiredVotingPower,
+		requiredCount:       requiredCount,
+		signatures:          make([][]byte, 0, s.Size()),
+		done:                make(chan struct{}),
+	}
+}
+
+// Add validates and adds a signature from the given validator.
+// Returns an error if the signature is invalid or the validator was already added.
```
cmwaters: I don't see where we check whether we've already validated that signature/validator

x/fibre/validator/signature_set.go
```diff
@@ -0,0 +1,117 @@
...
+	select {
+	case <-ss.done:
+	default:
+		close(ss.done)
+	}
+	return nil
+}
+
+// Miss marks the validator as not responding without adding a signature.
+func (ss *SignatureSet) Miss() {
```
cmwaters: How do you imagine this getting used? Don't we need to include which validator has missed?

x/fibre/validator/signature_set.go
```diff
@@ -0,0 +1,117 @@
...
+// Returns nil if thresholds are met or error if not enough signatures with voting power accumulated or
+// context errors.
+func (ss *SignatureSet) Await(ctx context.Context) error {
+	select {
+	case <-ss.done:
+		ss.mu.Lock()
+		defer ss.mu.Unlock()
+		if ss.votingPower < ss.requiredVotingPower {
+			return ErrNotEnoughSignatures
+		}
```
cmwaters: Why do we need this check? Can done be called when there isn't enough voting power?

x/fibre/validator/signature_set.go
Wondertan: https://github.com/celestiaorg/celestia-app-fibre/pull/6#discussion_r2430332017

Wondertan: This thread should answer your questions. On needing to know a particular validator: in one of the versions I had `core.Validator `being passed just in case, but in fact you end up not needing it here. The caller of this function already knows who the validator and can react accordingly. It will be much clearer once we get to `client.Put`

x/fibre/validator/signature_set.go
Wondertan: Basically if all or some validators end up with `Miss`, the done channel will unblock and then hit this case returning the error. Unblocking means we got sufficient validator responses(controlled by requiredCount) to make a final decision and the final decision is based on whether we got sufficient votingPower.

x/fibre/validator/signature_set.go
Wondertan: [A few lines below](https://github.com/celestiaorg/celestia-app-fibre/pull/6/files#diff-07d6a9dd982978665c9243e67212fd7622dc0a069e73ca6056273a6eb111a84aR52-R55)

x/fibre/validator/signature_set.go
```diff
@@ -0,0 +1,117 @@
...
+// Await waits for the signature collection thresholds to be met or for the context to be cancelled.
+// Returns nil if thresholds are met or error if not enough signatures with voting power accumulated or
+// context errors.
+func (ss *SignatureSet) Await(ctx context.Context) error {
+	select {
+	case <-ss.done:
+		ss.mu.Lock()
+		defer ss.mu.Unlock()
+		if ss.votingPower < ss.requiredVotingPower {
+			return ErrNotEnoughSignatures
```
rootulp: [no change needed][question] do we want two distinct errors to aid debugging:
1. ErrNotEnoughSignatures
2. ErrNotEnoughVotingPower

x/fibre/validator/signature_set.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	return nil
+}
+
+// Miss marks the validator as not responding without adding a signature.
+func (ss *SignatureSet) Miss() {
+	ss.mu.Lock()
+	defer ss.mu.Unlock()
+
+	ss.validatorCount++
```
rootulp: [question] for my own understanding, if a validator is not responsive why does a `Miss` add the validator to the signature set.

Intuitively, I would expect that an unresponsive validator means that the validator is offline and therefore didn't store a user's Fibre blob and therefore shouldn't count towards the threshold needed to satisfy the threshold for a signature set.

rootulp (dismissed):

x/fibre/validator/signature_set.go
Wondertan: Bingo

There is bug here that I realized after writing [this](https://github.com/celestiaorg/celestia-app-fibre/pull/6#discussion_r2432100671) response.

Basically, I coupled the requests with total count, so according to current logic it is possible for client to think that client.Put was unsuccessful too early without awaiting all the responses from all validators.

x/fibre/validator/signature_set.go
Wondertan: I think it's worth distinguishing, so I will add it.

Wondertan:
I made one more change, which removes `Miss` and `Await` methods and adds `Done` method. There was a bug described in [here](https://github.com/celestiaorg/celestia-app-fibre/pull/6#discussion_r2433654237). The solution to this bug is not to track the unblocking condition inside the previous `Await`, but on the caller side, thus the `Done` method that notifies when the thresholds are reached.


@evan-forbes, what I explained regarding not waiting for the timeout once we have all responses still stays true, it just handling of this is moved outside of signature set

x/fibre/validator/signature_set.go
Wondertan: So, to answer the question, we shouldn't track unresponsive validators as count, thus the fix in [48aafa1](https://github.com/celestiaorg/celestia-app-fibre/pull/6/commits/48aafa1ec133811683fe515ec864eefbdbd86acb)

x/fibre/validator/signature_set.go
Wondertan: I added the second error [48aafa1](https://github.com/celestiaorg/celestia-app-fibre/pull/6/commits/48aafa1ec133811683fe515ec864eefbdbd86acb)

x/fibre/validator/signature_set.go
```diff
@@ -0,0 +1,95 @@
...
+func (ss *SignatureSet) Signatures() ([][]byte, error) {
+	ss.mu.Lock()
+	defer ss.mu.Unlock()
+
+	if len(ss.signatures) < ss.minRequiredSignatures {
+		return nil, ErrNotEnoughSignatures
+	}
+	if ss.votingPower < ss.minRequiredVotingPower {
+		return nil, ErrNotEnoughVotingPower
+	}
```
walldiss: Would it make sense to allow returning signatures if not enough collected? This might be helpful for retries / resubmission to not sent data to validators that already saw failed blob.

x/fibre/validator/signature_set_test.go
```diff
@@ -0,0 +1,170 @@
...
+		for i := 0; i < 2; i++ {
+			signature, err := s.privKeys[i].Sign(s.signBytes)
+			require.NoError(t, err)
+			require.NoError(t, s.sigSet.Add(s.validators[i], signature))
+		}
+
+		sigs, err := s.sigSet.Signatures()
+		require.ErrorIs(t, err, validator.ErrNotEnoughSignatures)
+		require.Nil(t, sigs)
+	})
```
walldiss, line 116: Perhaps worth to check that done channel is still not closed (unfinished collection)

x/fibre/validator/signature_set.go
Wondertan: I dont think this is going to be immediately helpful looking at how the current client is structured.

Wondertan: As to retries, we can do them on grpc client level and if they dont work we fail.

x/fibre/validator/signature_set_test.go
Wondertan: Another test covers this

x/fibre/validator/signature_set.go
walldiss: On failed case, do you plan to return collected signatures to the caller? I think it is valuable information

Wondertan: It might become valuable in the future, but its not currently. Once it becomes we can add a new error type that collects partial signature set to caller.

Wondertan: I added this, but you owe me now two unnecessary changes

walldiss (approved):

x/fibre/validator/signature_set.go
walldiss: I’m trying to keep reviews quick and constructive. When you disagree, please share the rationale so we can converge faster. Phrasing like “you owe me” isn’t professional. Let’s stick to technical arguments and we’ll resolve this efficiently.

PR #4, opened 2025-10-12
feat(x/fibre): initial keeper implementation

rootulp:
Part of https://github.com/celestiaorg/celestia-app-fibre/issues/1

Adds basic CRUD operations to the keeper for:
- Escrow accounts (Get, Set, Delete)
- Withdrawals (Get, Set, Delete) and GetBySigner
- ProcessedPayment (Get, Set, Delete) and IsPaymentProcessed

We may need more later 🤷‍♂️ but this seems safe to merge as-is

PR #8, opened 2025-10-13
feat(x/fibre): add service proto

Wondertan:
~The GRPC service definition was added outside the centralized proto definition in `x/fibre/proto`, as the fibre service doesn't need to follow the cosmos framework, nor is it beneficial to mix unrelated things in a single proto dump. Consequently, the proto-gen script was updated to support this mode.~

The above is no longer true and the service protos are now in the central proto dir.

The PaymentPromise definition had its fields reordered and regrouped based on logical proximity (Also, row_version was renamed to blob_version). Looking at the #4 PR now, a similar change was made there as well.

Notable divergencies from the spec:
* `Row` instead of `RowWithProof` with an optional `rlc_root` field that is used only for the retrieval path. 
* `GetRowsResponse` without RlcOrig as inclusion verification on retrieval doesn't need the full RlcOrig
* Removes `TTL` from responses, as the client side can compute it itself
* Removes `Backoff`  as currently unused (may not even be needed as it is a transport concern rather than a model)
* Removes the `Commitment` message as it can be a simple field that is used once.
* Inlines `RlrOrig` message into `UploadRowsRequest`
* Renames `GetRowsRequest` to `DownloadRowsRequest`


x/fibre/proto/fibre.proto
```diff
@@ -0,0 +1,41 @@
+syntax = "proto3";
+package fibre.service.v1;
+
+option go_package = "github.com/celestiaorg/celestia-app/v6/x/fibre/proto";
```
evan-forbes: > as the fibre service doesn't need to follow the cosmos framework, nor is it beneficial to mix unrelated things in a single proto dump.

I'm fine with keeping things here, but seems better to use the existing proto package in celestia-app/proto since all the infra is there and just not follow the types thing

proto/celestia/fibre/v1/fibre.proto
```diff
@@ -27,27 +27,27 @@ message EscrowAccount {
 // PaymentPromise is a promise to pay for a fibre blob. It contains the
 // commitment and payment details for the fibre blob.
 message PaymentPromise {
-  // signer_public_key is the public key of the owner of the escrow account to charge
-  google.protobuf.Any signer_public_key = 1 [(cosmos_proto.accepts_interface) = "cosmos.crypto.PubKey"];
+  // chain_id is the chain ID that this payment promise is valid for.
```
evan-forbes, line 30: not blocking but why move this?

evan-forbes (approved):

x/fibre/proto/fibre.proto
evan-forbes: I get not wanting to follow the sdk's flawed design, but having duplicate infra seems likely completely avoidable tech debt. What are the benefits? tbh I initially thought it was just AI slop before reading your comment to find it was on purpose.

proto/celestia/fibre/v1/fibre.proto
Wondertan: > The PaymentPromise definition had its fields reordered and regrouped based on logical proximity (Also, row_version was renamed to blob_version). Looking at the #4 PR now, a similar change was made there as well.

To group with signature field. Rootul intuitively did the same in #4.

evan-forbes: I read your comment, but why is this order more logical?

x/fibre/proto/fibre.proto
Wondertan: I always get lost in apps proto pkg and I am generally used to define protos right next to their usage in protocols.

Wondertan: I am not sure what do you mean by duplicated infra. Do you mean the addition of buf files? These are final and dont need anything else to change or add or improve, so dont they qualify as debt afaiu

evan-forbes: > There are final

maybe, but now we have two. why not do this for every package? its tech debt. I don't care about the code, tbh I only find it annoying because you were just preaching about "principled engineering" but then don't follow the oldest rule in the book because it doesn't suit you.

Wondertan: Generally, I don't feel strongly about modularizing protos. Its just the way I used to do things and its more of a proposal to start going this way.  However, if you and other app maintainers do not want to go this path, I am happy to revert.

evan-forbes: I think we should follow the patterns of the repo when it makes sense to keep things tidy. keeping protos in the protos directory and minimizing the scripts needed seems reasonable to me.

agree with the sentiment that we don't need to follow the sdk's organizational patterns with "types" etc for non-statemachine things

evan-forbes: thanks :pray:

Wondertan: >>There are final

> maybe, but now we have two. why not do this for every package? its tech debt. I don't care about the code, tbh I only find it annoying because you were just preaching about "principled engineering" but then don't follow the oldest rule in the book because it doesn't suit you.

Wow, there must have been a confusion. I didn't meant it like the changes in this PR are final as like I am not willing to do any changes or incorporate the feedback. I meant that the buf configurations are final, as in do you often need to change buf configuration? Its something you do once and then dont touch until they do some breaking changes a few times in lifetime.

evan-forbes: > Its something you do once and then dont touch until they do some breaking changes a few times in lifetime.

I get that, but we shouldn't have two here imo for the same reason as having one per package. The few times we do change it its a huge footgun. Its intuitive that if we change one config, the entire repo changes.

When following DRY doesn't make things implicit, imo we should do that. Simultaneously, if the code is organized a specific way, stick to that schema. If we want to change it, then we should change everything to keep things organized. 

I appreciate you being willing to incorporate feedback, mine is not blocking because this is not that important, but that is my feedback.

Wondertan: > Simultaneously, if the code is organized a specific way, stick to that schema. If we want to change it, then we should change everything to keep things organized.

I do agree with this! If I had to add new protos for something reactor specific or new sdk module I would follow the existing pattern or refactor everything together with a new one. 

This case on the other hand feels different, as to me its neither of what cosmos protos were made and patterns set up to. If I didn't see it as a special case, I would pull PP here too, but its kept in central proto in this PR.

Note that I am not arguing back and simply want you to understand my intuition here.

x/fibre/proto/fibre.proto
```diff
@@ -0,0 +1,41 @@
...
+
+import "celestia/fibre/v1/fibre.proto";
+
+message Row {
+  uint32 index = 1;
+  bytes data = 2;
+  repeated bytes proof = 3;
+  optional bytes rlc_root = 4;                            // len == 32 when present
```
walldiss: Why rlc_root is not part of the proof?

x/fibre/proto/fibre.proto
```diff
@@ -0,0 +1,41 @@
...
+package fibre.service.v1;
+
+option go_package = "github.com/celestiaorg/celestia-app/v6/x/fibre/proto";
+
+import "celestia/fibre/v1/fibre.proto";
+
+message Row {
+  uint32 index = 1;
+  bytes data = 2;
+  repeated bytes proof = 3;
```
walldiss: What is encoding schema of proof? Does it contain fields that can be proto encoded?

I know it was raw bytes in spec, but it was because proof structure was undefined from rsem1d and we should do proper Proof message instead.

walldiss (commented):
> Removes Backoff as currently unused (may not even be needed as it is a transport concern rather than a model)

Backoff has nothing to do with transport. It is meant for backpressure. I'm fine to remove it for now, created issue to revisit it later. 

---
file names makes it hard to digest and find content:
- proto/celestia/fibre/v1/fibre.proto
- x/fibre/proto/fibre.proto

It is just fibre fibre fibre. I think `service.proto` and `payments.proto` would make more sense. And grouping them into single folder

Wondertan:
> Backoff has nothing to do with transport. It is meant for backpressure. I'm fine to remove it for now, created issue to revisit it later.

By transport concern I mean that if it can be done on a transport level rather then application, then it should. The flow control/backpressure is part of QUIC stream functionality for example. If its not possible with GRPC, then ok

x/fibre/proto/fibre.proto
Wondertan: Because we don't need it when we do UploadRowsRequest, as we send the whole RlcOrig.

Wondertan: One thing I start to wonder if really need to send RlcOrig ever. We could just send multiproof to rows and rlc, which presumable should be less then sending the whole rlcOrig.

x/fibre/proto/fibre.proto
Wondertan: Why should we extract Proof into a standalone message?

Wondertan: > What is encoding schema of proof? Does it contain fields that can be proto encoded?

It's a 2d array of hashes

Wondertan: Similarly, there is GF128 message which could be reduced to `repeated bytes` instead or kept as a separate message.

x/fibre/proto/fibre.proto
Wondertan: Move to central proto dir.

x/fibre/proto/fibre.proto
walldiss: We should send rlc_orig because server need to verify full rlc and not only inclusion. 

Perhaps you didn't get the question about proof. The rcl_root is part of the inclusion proof to commitment, it is one of the proof nodes. It does not need to be send as separate item

x/fibre/proto/fibre.proto
walldiss: Makes sense. my confusion came from the fact that there is needs to be index, but it already part of row

x/fibre/proto/fibre.proto
Wondertan: > The rcl_root is part of the inclusion proof to commitment,

It is not as per the current implementation of proof! I have a change locally that add a new InclusionProof type to rsema which includes the rlc_root.

Wondertan: [Ref](https://github.com/celestiaorg/rsema1d/blob/65cc9d4c3887936393763d32e0cf4e665d9d2705/codec.go#L108-L116)

walldiss (approved):

proto/celestia/fibre/v1/fibre.proto
```diff
@@ -27,27 +27,27 @@ message EscrowAccount {
...
+  // height is the height that is used to determine the validator set that is used
+  int64 height = 2;
   // namespace is the namespace the blob is associated with.
-  bytes namespace = 2;
+  bytes namespace = 3;
   // blob_size is the size of the blob in bytes
-  uint32 blob_size = 3;
+  uint32 blob_size = 4;
+  // blob_version is the version of the blob format
+  uint32 blob_version = 5;
```
cmwaters, line 40: nit: if we are going to change it from row version to something I would prefer `fibre_version` to keep it distinct from any existing notion around blobs. 

Fibre version is mostly about the off-chain communication protocol between client and server (and that includes the row encoding schema)

rootulp: `fibre_blob_version` was the name proposal in https://github.com/celestiaorg/fibre-da-spec/issues/20

PR #9, opened 2025-10-13
feat(x/fibre): add `HostRegistry` interface 

Wondertan:
`HostRegistry` provides a mapping from validators to their network hosts, enabling the fibre client to resolve network addresses for validator connections.

The interface will be implemented once https://github.com/celestiaorg/fibre-da-spec/pull/4 is done. There might also be auxiliary testing implementations with a file-based registry or similar, rather than a state tree.

The interface currently returns just the host string. In an alternative design, it could return a GRPC client connection given the respective validator. However, the decision was not to rush coupling the transport layer and the state registry yet.

### DIFF
--- a/x/fibre/validator/host.go
+++ b/x/fibre/validator/host.go
@@ -1,0 +1,23 @@
+package validator
+
+import (
+	"context"
+
+	core "github.com/cometbft/cometbft/types"
+)
+
+// Host represents a validator's network address as a string.
+// It can be a DNS name, IPv4 address, or IPv6 address, optionally with a port.
+type Host string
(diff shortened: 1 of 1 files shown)


walldiss (approved):
ltgm

cmwaters (commented):
FYI, this might change depending on the query methods exposed by the x/valaddr module

Wondertan:
Iirc, the query has no parameters and just fetches everything. The core.Validator struct has validator address that we can key hosts by internally or even externally if end up adding a query for granular validator host retrieval

PR #11, opened 2025-10-14
feat(x/fibre): Add `Assign` method to `validator.Set`

Wondertan:
~This diverges from the spec by not being called or abstracted away as a `ShardMap`. In practice, we won't have multiple algorithms implemented simultaneously and new ones will simply replace the older ones. 
This is also implemented as a method on `Set` structure rather then a separate component. A separate component is unnecessary as there is no state that the component would maintain and assigning is a pure function. Once we decide on the final assigning, we can simply rewrite the method as needed and update the respective usage code.~

The above is no longer true. The `ShardMap` type was added which holds validator to indexes assignments as per spec. Originally the idea was to individually select validators, but as it turns out it is impossible thus we need add the whole map.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -59,8 +59,6 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
-require github.com/celestiaorg/reedsolomon v1.12.6-0.20250824224240-8b66bda83fd0 // indirect
-
 require (
 	cel.dev/expr v0.24.0 // indirect
 	cloud.google.com/go v0.120.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -59,6 +59,8 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
+require github.com/celestiaorg/reedsolomon v1.12.6-0.20250824224240-8b66bda83fd0 // indirect
+
 require (
 	cel.dev/expr v0.24.0 // indirect
 	cloud.google.com/go v0.120.0 // indirect
--- a/x/fibre/validator/set.go
+++ b/x/fibre/validator/set.go
@@ -1,6 +1,9 @@
 package validator
 
 import (
+	"math/big"
+
+	"github.com/celestiaorg/rsema1d"
 	core "github.com/cometbft/cometbft/types"
 )
 
@@ -8,4 +11,28 @@ import (
 type Set struct {
 	*core.ValidatorSet
 	Height uint64
+}
+
+// Assign returns a validator for the given commitment and row index using the formula:
+//
+//	validator_index = (commitment + row_index) mod num_validators
+//
+// The commitment is converted to a big.Int, added to the row index, and the result
+// is taken modulo the number of validators to determine the assignment.
+//
+// TODO(@Wondertan): This assignment algorithm is not final and may be changed
+// to improve distribution properties or security guarantees.
+func (s Set) Assign(commitment rsema1d.Commitment, rowIndex int) *core.Validator {
+	if len(s.Validators) == 0 {
+		return nil
+	}
+
+	// TODO(@Wondertan): If we ever end up using this assignment algorithm,
+	// we should move to uint256 libraries for up to 60% speedups per arithmetic operations
+	commitmentInt := new(big.Int).SetBytes(commitment[:])
+	rowIndexInt := big.NewInt(int64(rowIndex))
+	sum := new(big.Int).Add(commitmentInt, rowIndexInt)
+	valLenBig := big.NewInt(int64(len(s.Validators)))
+	idx := new(big.Int).Mod(sum, valLenBig)
+	return s.Validators[idx.Int64()]
 }
--- a/x/fibre/validator/set.go
+++ b/x/fibre/validator/set.go
@@ -13,7 +13,11 @@ type Set struct {
 	Height uint64
 }
 
-// Assign returns a validator for the given commitment and row index using the formula:
+// ShardMap maps validators to the row indices they are assigned.
+type ShardMap map[*core.Validator][]int
+
+// Assign returns a ShardMap containing all validators and their assigned row indices
+// for the given commitment and total number of rows. The assignment uses the formula:
 //
 //	validator_index = (commitment + row_index) mod num_validators
 //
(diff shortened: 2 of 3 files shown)


cmwaters (commented):
Why do you want to have this in x/fibre. I had thought that x/fibre would be for all state machine related code as it is where the module is hosted and is the convention for other sdk module. I would have thought that we would have a different structure like

--> x/fibre
--> fibre
----->client
----->server
----->common

^ this is just an example to show the separation.

The state machine does not have to be aware of the row dispersion algorithm so it should reside in `common` or something similar

Wondertan:
I am in favor of moving fibre stuff unrelated to statemachine outside of `x/fibre`. However, if we do it, the proto definitions for grpc client and server would have to still be in `x/fibre/types`, unless @evan-forbes permits to split those apart.

walldiss:
`ShardMap` does not have to component, but just function type. The goal is that we can replace the function for easier testing and imply constant dispersion instead of sharding algo for debugging.

x/fibre/validator/set.go
```diff
@@ -9,3 +12,27 @@ type Set struct {
...
+
+	// TODO(@Wondertan): If we ever end up using this assignment algorithm,
+	// we should move to uint256 libraries for up to 60% speedups per arithmetic operations
+	commitmentInt := new(big.Int).SetBytes(commitment[:])
+	rowIndexInt := big.NewInt(int64(rowIndex))
+	sum := new(big.Int).Add(commitmentInt, rowIndexInt)
+	valLenBig := big.NewInt(int64(len(s.Validators)))
+	idx := new(big.Int).Mod(sum, valLenBig)
+	return s.Validators[idx.Int64()]
+}
```
walldiss: This approach implies we would need to run this funciton against every row index. Which seems intuitivly not optimal. To ensure function is optimal it should not take rowIndex as an input. But rather take commitment and whole set and return map of assignments. If algo is based on shuffle it will work much faster then iterative approach

What would be assimptotic complexity of this algo?

x/fibre/validator/set_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"testing"
+
+	"github.com/celestiaorg/celestia-app/v6/x/fibre/validator"
+	"github.com/celestiaorg/rsema1d"
+	"github.com/cometbft/cometbft/crypto/ed25519"
+	core "github.com/cometbft/cometbft/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSet_Assign(t *testing.T) {
```
walldiss, line 14: Would be nice to add simple benchmark to see how fast assignemt is calculated for 100 validators.

Wondertan: I actually had it, will upload. Its much faster then then anything we proposed so far

x/fibre/validator/set.go
Wondertan: Well, it just O(n) where n is number of rows, which is very simple and why I like it. The problem with it is that it might not be as random and that's why its not final, but can be easily replaced.

walldiss: Ah, this is just kinda mock to have some function to distribute it. I'm not opposed to this algo for testing, but I think the Assign func signature should return map so we can use replace it easily with proper random shuffle algo later

evan-forbes:
> I am in favor of moving fibre stuff unrelated to statemachine outside of x/fibre

> However, if we do it, the proto definitions for grpc client and server would have to still be in x/fibre/types

we can definitely split them up. the entire point that I cared about for the protos was to just keep the protos in a single spot so we don't have to duplicate configs or scripts. not that we keep the cosmos-sdk "types" package for all types. for the state machine, it makes sense to keep that. for the non-statemachine we don't need to. This is what we do for all of our non-state machine protos see https://github.com/celestiaorg/celestia-app/tree/main/proto/celestia/core/v1 . protos are there, yet they don't follow the types package

go.mod
```diff
@@ -59,6 +59,8 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
+require github.com/celestiaorg/reedsolomon v1.12.6-0.20250824224240-8b66bda83fd0 // indirect
```
rootulp: [uber nit] odd that this didn't get grouped with the other indirect dependencies in the block below. I think we could manually move it into the `require` block below

Wondertan:
> This is what we do for all of our non-state machine protos see https://github.com/celestiaorg/celestia-app/tree/main/proto/celestia/core/v1 . protos are there, yet they don't follow the types package

These are still in the `proto` and not actually grouped with code using them. In this case it doesn't make any difference whether they are in `core` or `fibre` subfolder. Furthermore it would make it seem like fibre grpc service is part of core when it is not.

>  the entire point that I cared about for the protos was to just keep the protos in a single spot so we don't have to duplicate configs or scripts

There is no way to split without changes to buf configuration or scripts and I tried it.

rootulp (dismissed):
Agreed w/ Callum that this seems possible to extract from x/fibre because it's not needed for the state machine

x/fibre/validator/set_test.go
Wondertan: Done

x/fibre/validator/set.go
Wondertan: As per sync discussion, we decided to go with the map, as there we can't have individually assigned validators and they must be always grouped with a truly random algo

go.mod
Wondertan: resolved by rebase

nashqueue:
Edit: My comment was on a previous state of the PR, which seems to be already resolved, but I will leave it up anyway. 

Because we do non-overlapping shares, we have the following calculation: We have r rows and v validators, so every validator should get at least q = ⌊r/v⌋ shares. And some get q+1 

If you want a pure function to calculate the mapping that can be replaced later, it has to apply it to all rows and validators. We should not assume that we can do it for a particular validator or index. So giving an index and getting a validator is a special case for this algorithm. 

It should be the other way around, where you give a validator and you get the list of row indices for that validator. So I guess it would be more like (commitment, Vset, validator) -> (List[row_i]) 

Underneath, it could either perform your algorithm or the one the client uses as well, which would apply to all rows and validators.  The input is (commitment, Vset), and the output is a list of row indices [row_i].    The first q rows of this output list go to the first validator, the next to the second, and so on.

walldiss (approved):

PR #12, opened 2025-10-14
feat(x/fibre): add `PaymentPromise` type

Wondertan:
Adds a non-generated `PaymentPromise` type in the fibre package that wraps the proto definition and provides additional methods. This allows extending the type with caching for fields that would be heap allocated anyway (sign bytes and hash), as well as serialization helpers and validation logic.

Then, the proto of `PaymentPromise` is updated again to use respective public key proto type available in cosmos proto.

Besides, makes fibre service to be generated in `types` pkg. This proto framework enforces everything to be placed in a single place and even generated in a single place and there is no way around this. Enforcing everything to be in a single pkg is misguided, shortsighted and simply dumb and I am exhausted from fighting this fight with the proto setup without changing it and I surrender. It makes me wanna rage quit this repo... I blame Jae Kwon for this, although he is unlikely the reason...


proto/celestia/fibre/v1/fibre.proto
```diff
@@ -45,7 +45,7 @@ message PaymentPromise {
   // determining when service stops for this blob.
   google.protobuf.Timestamp creation_timestamp = 7 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
   // signer_public_key is the public key of the owner of the escrow account to charge
-  google.protobuf.Any signer_public_key = 8 [(cosmos_proto.accepts_interface) = "cosmos.crypto.PubKey"];
+  cosmos.crypto.secp256k1.PubKey signer_public_key = 8 [(gogoproto.nullable) = false];
```
evan-forbes, line 48: dope thanks

evan-forbes (approved):

Wondertan:
Wait, I have unrelated changes here somehow... 

I will force push them away from feature/fibre and make a new PR.

Here it goes #13

proto/celestia/fibre/v1/fibre.proto
cmwaters, line 48: Why are we enforcing secp256k1 when the cosmos-sdk supports other key formats?

Wondertan: I would want to support more but in fact keyring supports only this one.

rootulp: Good point. I thought Cosmos SDK also supported more key formats so I (or more accurately Cursor) used google.protobuf.Any for the signer_public_key to copy the implementation used in a few other places in the Cosmos SDK: 

- https://github.com/rootulp/cosmos-sdk/blob/fc6040622be5c50d713d72d8f5f3d6bc3e186c80/proto/cosmos/auth/v1beta1/auth.proto#L23
- https://github.com/rootulp/ibc-go/blob/a17d4368090a6ab6d97c04a1b3e45c21fd0b7478/proto/ibc/lightclients/solomachine/v3/solomachine.proto#L27

but Hlib's right that the keyring appears to only support one curve:

```
$ celestia-appd keys list-key-types
Supported key types/algos:
"secp256k1"
```

PR #13, opened 2025-10-14
feat(x/fibre/types): grpc client cache and ctor

Wondertan:
We need to cache/pool grpc clients and connections on the client side, so we don't reconnect all the time. The cache doesn't have any GC yet for when there are too many validator connections. 

Besides, a new function type was introduced that constructs the client and connection which is useful for mocking. The default one constructs insecure connection with support for OTLP traces that client and server integrate from day 0.

The insecure connection is something that we should not forget to change, until its too late. TLS + self signed certificates, unless we want to depend on centralized certificate authorities... (Also worth linking: https://github.com/open-telemetry/opentelemetry-go/issues/6661)

Also, these do not need to be part of `types` pkg but we sort of have to in order to avoid cycles

### DIFF
--- a/x/fibre/types/client_cache.go
+++ b/x/fibre/types/client_cache.go
@@ -1,0 +1,70 @@
+package types
+
+import (
+	"context"
+	"errors"
+	"sync"
+
+	core "github.com/cometbft/cometbft/types"
+)
+
+// ClientCache caches [FibreClientCloser]s per validator using the provided constructor function.
+// TODO(@Wondertan): Needs cleanup strategy, e.g. LRU
+type ClientCache struct {
+	newClient FibreClientCloserFn
+	mu        sync.Mutex
+	clients   map[string]*clientEntry // keyed by validator address string
+}
+
+// clientEntry holds a lazily-initialized [FibreClientCloser].
+type clientEntry struct {
+	sync.Once
+	clientCloser FibreClientCloser
+	err          error
+}
+
+// NewClientCache creates a new [ClientCache] with the given [FibreClientCloserFn].
+func NewClientCache(newClient FibreClientCloserFn) *ClientCache {
+	return &ClientCache{
+		newClient: newClient,
+		clients:   make(map[string]*clientEntry),
+	}
+}
+
+// GetClient returns a cached [FibreClient] for the validator, creating one if needed.
+// Uses the constructor function provided to [NewClientCache]. Only one dial per validator will occur.
+func (cc *ClientCache) GetClient(ctx context.Context, val *core.Validator) (FibreClient, error) {
+	addr := val.Address.String()
+
+	cc.mu.Lock()
+	entry, ok := cc.clients[addr]
+	if !ok {
+		entry = &clientEntry{}
+		cc.clients[addr] = entry
+	}
+	cc.mu.Unlock()
+
+	entry.Do(func() {
+		client, err := cc.newClient(ctx, val)
+		if err != nil {
+			entry.err = err
+			return
+		}
+		entry.clientCloser = client
+	})
+
+	return entry.clientCloser, entry.err
+}
+
+// Close closes all cached [FibreClientCloser]s.
+func (cc *ClientCache) Close() (err error) {
+	cc.mu.Lock()
+	defer cc.mu.Unlock()
+	for _, entry := range cc.clients {
+		if entry.clientCloser != nil {
+			err = errors.Join(err, entry.clientCloser.Close())
+		}
+	}
+	cc.clients = make(map[string]*clientEntry)
+	return err
+}
--- a/x/fibre/types/client_cache.go
+++ b/x/fibre/types/client_cache.go
@@ -8,23 +8,23 @@ import (
 	core "github.com/cometbft/cometbft/types"
 )
 
-// ClientCache caches [FibreClientCloser]s per validator using the provided constructor function.
+// ClientCache caches [Client]s per validator using the provided constructor function.
 // TODO(@Wondertan): Needs cleanup strategy, e.g. LRU
 type ClientCache struct {
-	newClient FibreClientCloserFn
+	newClient NewClientFn
 	mu        sync.Mutex
 	clients   map[string]*clientEntry // keyed by validator address string
 }
 
-// clientEntry holds a lazily-initialized [FibreClientCloser].
+// clientEntry holds a lazily-initialized [Client].
 type clientEntry struct {
 	sync.Once
-	clientCloser FibreClientCloser
+	clientCloser Client
 	err          error
 }
 
-// NewClientCache creates a new [ClientCache] with the given [FibreClientCloserFn].
-func NewClientCache(newClient FibreClientCloserFn) *ClientCache {
+// NewClientCache creates a new [ClientCache] with the given [NewClientFn].
+func NewClientCache(newClient NewClientFn) *ClientCache {
 	return &ClientCache{
 		newClient: newClient,
 		clients:   make(map[string]*clientEntry),
@@ -56,7 +56,7 @@ func (cc *ClientCache) GetClient(ctx context.Context, val *core.Validator) (Fibr
 	return entry.clientCloser, entry.err
 }
 
-// Close closes all cached [FibreClientCloser]s.
+// Close closes all cached [Client]s.
 func (cc *ClientCache) Close() (err error) {
 	cc.mu.Lock()
 	defer cc.mu.Unlock()
--- a/x/fibre/types/fibre_client.go
+++ b/x/fibre/types/fibre_client.go
@@ -1,0 +1,59 @@
+package types
+
+import (
+	"context"
+	"io"
+
+	"github.com/celestiaorg/celestia-app/v6/x/fibre/validator"
+	core "github.com/cometbft/cometbft/types"
+	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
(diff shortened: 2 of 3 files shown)


evan-forbes (dismissed):

evan-forbes:
Mind opening short issues to ensure we don't forget the flups on the GC and the tls?

x/fibre/types/client_cache.go
```diff
@@ -0,0 +1,70 @@
...
+	"errors"
+	"sync"
+
+	core "github.com/cometbft/cometbft/types"
+)
+
+// ClientCache caches [FibreClientCloser]s per validator using the provided constructor function.
+// TODO(@Wondertan): Needs cleanup strategy, e.g. LRU
+type ClientCache struct {
+	newClient FibreClientCloserFn
```
walldiss: why closer is has prefix "new"? Seems wierd and misleading. Is it closing or creating something?

x/fibre/types/client_cache.go
```diff
@@ -0,0 +1,70 @@
...
+	mu        sync.Mutex
+	clients   map[string]*clientEntry // keyed by validator address string
+}
+
+// clientEntry holds a lazily-initialized [FibreClientCloser].
+type clientEntry struct {
+	sync.Once
+	clientCloser FibreClientCloser
+	err          error
+}
```
walldiss, line 24: Does it miss actual connection field?

x/fibre/types/client_cache.go
```diff
@@ -0,0 +1,70 @@
...
+
+// NewClientCache creates a new [ClientCache] with the given [FibreClientCloserFn].
+func NewClientCache(newClient FibreClientCloserFn) *ClientCache {
+	return &ClientCache{
+		newClient: newClient,
+		clients:   make(map[string]*clientEntry),
+	}
+}
```
walldiss: Nit: could prealloc map if size is passed to constructor. We know the size will be 100

x/fibre/types/client_cache.go
```diff
@@ -0,0 +1,70 @@
...
+func (cc *ClientCache) GetClient(ctx context.Context, val *core.Validator) (FibreClient, error) {
+	addr := val.Address.String()
+
+	cc.mu.Lock()
+	entry, ok := cc.clients[addr]
+	if !ok {
+		entry = &clientEntry{}
+		cc.clients[addr] = entry
+	}
+	cc.mu.Unlock()
```
walldiss, line 45: Should it fast return from cache if entry exists?

x/fibre/types/client_cache.go
```diff
@@ -0,0 +1,70 @@
...
+func (cc *ClientCache) Close() (err error) {
+	cc.mu.Lock()
+	defer cc.mu.Unlock()
+	for _, entry := range cc.clients {
+		if entry.clientCloser != nil {
+			err = errors.Join(err, entry.clientCloser.Close())
+		}
+	}
+	cc.clients = make(map[string]*clientEntry)
+	return err
```
walldiss, line 69: Don't need to keep lock on mutex whie calling Close(). It can take lock for extended time and prevent new connections from being opened

x/fibre/types/client_cache.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	cc.mu.Unlock()
+
+	entry.Do(func() {
+		client, err := cc.newClient(ctx, val)
+		if err != nil {
+			entry.err = err
+			return
+		}
+		entry.clientCloser = client
```
walldiss, line 53: what is the point of do once? We need to open only when entry is new and does not exist. So simple == nil check should be enough if you create connection before storing it into cache

x/fibre/types/fibre_client.go
```diff
@@ -0,0 +1,59 @@
...
+// FibreClientCloser combines [FibreClient] with [io.Closer] to manage the lifecycle
+// of both the client and its underlying connection.
+type FibreClientCloser interface {
+	FibreClient
+	io.Closer
+}
+
+// FibreClientCloserFn is a constructor function that creates a [FibreClientCloser]
+// for a given validator. It should handle host resolution and connection establishment.
+type FibreClientCloserFn func(ctx context.Context, val *core.Validator) (FibreClientCloser, error)
```
walldiss: Can we just store connections in cache and avoid unnecessory types abstractions?

walldiss (commented):
Seems a bit more complicated then needed for the purpose, but I might miss some edge cases that require all of extra abstractions and concurrency control logic. Would need to revisit on sync and discuss

x/fibre/types/client_cache.go
Wondertan: It is not Closer but CloserFn which may be new.

x/fibre/types/client_cache.go
Wondertan: Well we dont want to open new connections when the entire thing is being closed, we should be sropping from opening new connections in fact

x/fibre/types/client_cache.go
Wondertan: Because multiple client.Puts may ask for the same client

x/fibre/types/fibre_client.go
Wondertan: As in the PR description, there has to be some abstraction to mock the client away. If we store the connection on the cache, we cant mock the networking part away. The connection is a concrete type

x/fibre/types/client_cache.go
Wondertan: No, see the last comment on why

Wondertan:
> Would need to revisit on sync and discuss

This PR blocks me from upstreaming the client and we should aim to not hold it on sync discussions.


To reiterate, the abstraction is used for mocking and concurrency to support simultaneous Puts. If you find a way to simplify the grpc client lifecycle later on, this would be great, but lets please not block on this right now.

x/fibre/types/client_cache.go
Wondertan: We dont know the size of validator set her unless we pass it somehow through the config, but one more config doesnt seem to be worth it

x/fibre/types/client_cache.go
walldiss: We don't need to wait for it to be closed before opening new one. What is the reason to block here?

Wondertan: I mean new connections should not be opened at all by ClientCache when its being closed and afterwards, simply because its becoming closed for usage and thus new connections.

x/fibre/types/client_cache.go
Wondertan: In case of concurrent `GetClient` calls where both routines hit `entry.Do` and one of them aquires internal sync.Once lock, we need to ensure that the other one await until Do is finished by the former one. Otherwise, a nil client will be returned

x/fibre/types/client_cache.go
walldiss: New connection can be opened while old one is being closed. Why it should it wait for old one?

x/fibre/types/client_cache.go
Wondertan: I am open to better naming suggestions

x/fibre/types/fibre_client.go
walldiss: We should not mock grpc.Conn itself. Better to moch actual client with methods. proto will generate interface of service to mock against. If it is easier for you to reason around mocking we can keep grpc conn for now, but I'm sure you will see it is easier to mock with grpc.client interface

x/fibre/types/client_cache.go
walldiss: I see now. My confusion comes from thinking that clientClosr is `client closer`.  While in your case it is `Client and client closer` which means it is also a client. So I would suggest (non-blocking) to rename:

`FibreClientCloserFn` -> `NewClientFn`
`FibreClientCloser` -> `Client`

x/fibre/types/client_cache.go
walldiss: got it, got confused by naming

x/fibre/types/client_cache.go
walldiss: Still unresolved, but I think it can return fast if client is inited before

Wondertan: The client is inialized lazily so all the concurrent routines need to await until its done.

x/fibre/types/fibre_client.go
Wondertan: The `FibreClientCloser` is the client + closer, which mocks the actual client as you say. We don't mock the connection. The connection is hidden behind the `FibreClientCloser` particularly the `Closer` part, as we have to close them somehow still.

x/fibre/types/client_cache.go
Wondertan: Because when closing the ClientCache means closing it from further usage. I can add error boolean and return error if it helps.

walldiss: I mean you keep mutex lock whole time while connection is being closed. And I suggest to 
1. lock mutex 
2. take conn and delete if from map
3. unlock mutex
4. close connection

walldiss: nvm, I thought it closed individual connection. It would make sense if there was `Delete` method

Wondertan:
@evan-forbes, added the flups celestiaorg/celestia-app#6822 and celestiaorg/celestia-app#6821. 

I wish we could resolve them now, but celestiaorg/celestia-app#6822 is blocked on otlp and celestiaorg/celestia-app#6821 needs some more discussions

walldiss (approved):

rootulp (approved):

PR #14, opened 2025-10-14
feat(x/fibre): add `Blob` type (upload side)

Wondertan:
Add Blob type with methods and logic relevant for upload. The download side will come later once the Put side of the client is upstreamed. This is meant to simplify reviewing.

The tests covering the logic here do also exist, but they are convoluted with download logic and splitting them is to big of a lift, so I will add them together with download logic of the blob.

### DIFF
--- a/x/fibre/blob.go
+++ b/x/fibre/blob.go
@@ -11,10 +11,8 @@ import (
 	"github.com/celestiaorg/rsema1d/field"
 )
 
-var (
-	// ErrBlobTooLarge is returned when the blob size exceeds MaxBlobSize.
-	ErrBlobTooLarge = errors.New("blob size exceeds maximum allowed size")
-)
+// ErrBlobTooLarge is returned when the blob size exceeds MaxBlobSize.
+var ErrBlobTooLarge = errors.New("blob size exceeds maximum allowed size")
 
 // Commitment is a commitment to a blob.
 // TODO(@Wondertan): merge with rsema1d.Commitment once it has these methods.
@@ -112,7 +110,7 @@ func NewBlob(data []byte, cfg BlobConfig) (d *Blob, err error) {
 
 // Commitment returns the commitment to the blob.
 func (d *Blob) Commitment() Commitment {
-	return Commitment(d.commitment)
+	return d.commitment
 }
 
 // RLCOrig returns the original RLC coefficients.
--- a/x/fibre/blob.go
+++ b/x/fibre/blob.go
@@ -5,18 +5,19 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"math"
 	"runtime"
 
 	"github.com/celestiaorg/rsema1d"
 	"github.com/celestiaorg/rsema1d/field"
 )
 
-// ErrBlobTooLarge is returned when the blob size exceeds MaxBlobSize.
-var ErrBlobTooLarge = errors.New("blob size exceeds maximum allowed size")
+var (
+	// ErrBlobTooLarge is returned when the blob size exceeds MaxBlobSize.
+	ErrBlobTooLarge = errors.New("blob size exceeds maximum allowed size")
+)
 
-// Commitment is a commitment to fibre [Blob].
-// TODO(@Wondertan): merge with rsema1d.Commitment and move these methods.
+// Commitment is a commitment to a blob.
+// TODO(@Wondertan): merge with rsema1d.Commitment once it has these methods.
 type Commitment rsema1d.Commitment
 
 // UnmarshalBinary decodes a [Commitment] from bytes.
@@ -47,7 +48,7 @@ type BlobConfig struct {
 	// MaxBlobSize is the maximum allowed blob size.
 	MaxBlobSize int
 	// BlobVersion is the version of the row format.
-	BlobVersion uint32
+	BlobVersion uint8
 	// CodingWorkers is the number of workers to use for encoding and decoding rsema1d.
 	CodingWorkers int
 }
@@ -72,32 +73,34 @@ type Blob struct {
 	commitment   Commitment
 	rlcOrig      []field.GF128
 
-	// rows holds the shards for both encoded data and reconstruction.
-	rows [][]byte
+	// holds meta fields about the blob
+	header blobHeaderV0
+	// data holds the decoded original data (without header).
+	data []byte
 }
 
-// NewBlob creates a new [Blob] instance by encoding the original data.
-// It takes the original data and a [BlobConfig].
-// The data is prefixed with a header containing the blob version and original data size.
-func NewBlob(originalData []byte, cfg BlobConfig) (d *Blob, err error) {
-	if len(originalData) == 0 {
+// NewBlob creates a new [Blob] instance by encoding the data.
+// It takes the data and a [BlobConfig].
+// The data is prefixed with a header containing the blob version and data size.
+func NewBlob(data []byte, cfg BlobConfig) (d *Blob, err error) {
+	if len(data) == 0 {
 		return nil, fmt.Errorf("data cannot be empty")
 	}
-	if len(originalData) > cfg.MaxBlobSize {
-		return nil, fmt.Errorf("%w: data size %d exceeds maximum %d", ErrBlobTooLarge, len(originalData), cfg.MaxBlobSize)
+	if len(data) > cfg.MaxBlobSize {
+		return nil, fmt.Errorf("%w: data size %d exceeds maximum %d", ErrBlobTooLarge, len(data), cfg.MaxBlobSize)
 	}
 
 	d = &Blob{
-		cfg: cfg,
+		cfg:    cfg,
+		header: newBlobHeaderV0(len(data)),
+		data:   data,
 	}
 
-	rowSize := d.calculateRowSize(len(originalData))
-	d.rows = d.splitIntoRows(originalData, rowSize)
-
-	d.extendedData, d.commitment, d.rlcOrig, err = rsema1d.Encode(d.rows, &rsema1d.Config{
+	rows := d.header.encodeToRows(data, cfg)
+	d.extendedData, d.commitment, d.rlcOrig, err = rsema1d.Encode(rows, &rsema1d.Config{
 		K:           cfg.OriginalRows,
 		N:           cfg.ParityRows,
-		RowSize:     rowSize,
+		RowSize:     len(rows[0]),
 		WorkerCount: cfg.CodingWorkers,
 	})
 	if err != nil {
@@ -109,7 +112,7 @@ func NewBlob(originalData []byte, cfg BlobConfig) (d *Blob, err error) {
 
 // Commitment returns the commitment to the blob.
 func (d *Blob) Commitment() Commitment {
-	return d.commitment
+	return Commitment(d.commitment)
 }
 
 // RLCOrig returns the original RLC coefficients.
@@ -118,36 +121,35 @@ func (d *Blob) RLCOrig() []field.GF128 {
 }
 
 // RowSize returns the size of each row in bytes.
+// Returns 0 if no original data available to determine row size.
 func (d *Blob) RowSize() int {
-	if len(d.rows) > 0 && len(d.rows[0]) > 0 {
-		return len(d.rows[0])
+	if len(d.data) == 0 {
+		return 0
 	}
-	return 0
+
+	return d.header.calculateRowSize(len(d.data), d.cfg)
 }
 
 // DataSize returns the size of the original data (without header) by reading from the blob header.
-// Returns 0 if the header cannot be read.
+// Returns 0 if no original data available to determine its size.
 func (d *Blob) DataSize() int {
-	if len(d.rows) == 0 {
-		return 0
-	}
-
-	// extract size from header (first row)
-	if len(d.rows[0]) < blobHeaderSize {
-		return 0
-	}
-
-	return int(binary.BigEndian.Uint32(d.rows[0][math.MaxUint32 : math.MaxUint32*2]))
+	return len(d.data)
 }
 
 // Size returns the total size of the blob including the header overhead.
-// Returns 0 if the size cannot be determined.
+// Returns 0 if no original data available to determine blob size.
 func (d *Blob) Size() int {
 	dataSize := d.DataSize()
 	if dataSize == 0 {
 		return 0
 	}
 	return blobHeaderSize + dataSize
+}
+
+// Data returns the cached original data (without header).
+// Returns nil if the data has not been decoded yet (call Reconstruct first for received blobs).
+func (d *Blob) Data() []byte {
+	return d.data
 }
 
 // Row returns the [rsema1d.RowProof] for the given index from the extended data.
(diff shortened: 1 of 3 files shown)


evan-forbes (dismissed):
Might have missed them but ofc some unit tests could be nice

x/fibre/blob.go
```diff
@@ -0,0 +1,225 @@
...
+	// Total rows = OriginalRows + ParityRows.
+	ParityRows int
+	// RowSizeMin is the minimum row size in bytes.
+	RowSizeMin int
+	// MaxBlobSize is the maximum allowed blob size.
+	MaxBlobSize int
+	// BlobVersion is the version of the row format.
+	BlobVersion uint32
+	// CodingWorkers is the number of workers to use for encoding and decoding rsema1d.
+	CodingWorkers int
```
cmwaters, line 51: nit: call this `WorkerCount` to keep consistent with the term used in rsme1d

x/fibre/blob.go
```diff
@@ -0,0 +1,225 @@
...
+	// remaining rows: use slices from data (offset by what we already used)
+	dataOffset := firstRowDataSize
+	for i := 1; i < d.cfg.OriginalRows; i++ {
+		start := dataOffset
+		end := start + rowSize
+		dataOffset += rowSize
+
+		if end <= len(data) {
+			// full row available in data - use slice directly
+			rows[i] = data[start:end:end]
```
cmwaters, line 213: TIL

cmwaters (dismissed):
Logic looks sound. As Evan pointed out can you also add unit tests

Wondertan:
On unit tests

> The tests covering the logic here do also exist, but they are convoluted with download logic and splitting them is to big of a lift, so I will add them together with download logic of the blob.

x/fibre/blob.go
walldiss, line 51: What is the reason to make cfg public and making blob configurable? Most of config values must be enforced by protocol and should not allow user to alternate values.

x/fibre/blob.go
```diff
@@ -0,0 +1,225 @@
...
+	if len(data) != 32 {
+		return fmt.Errorf("commitment must be 32 bytes, got %d", len(data))
+	}
+	copy(c[:], data)
+	return nil
+}
+
+// String returns the hex-encoded string representation of the commitment.
+func (c Commitment) String() string {
+	return hex.EncodeToString(c[:])
```
walldiss, line 32: i might be wrong, but subslicing with `[:]` should not be necessary.

x/fibre/blob.go
```diff
@@ -0,0 +1,225 @@
...
+		return nil, fmt.Errorf("%w: data size %d exceeds maximum %d", ErrBlobTooLarge, len(originalData), cfg.MaxBlobSize)
+	}
+
+	d = &Blob{
+		cfg: cfg,
+	}
+
+	rowSize := d.calculateRowSize(len(originalData))
+	d.rows = d.splitIntoRows(originalData, rowSize)
+
```
walldiss: Is there a section where row and blob version are encoded into blob first 5 bytes? Seems like this encoding logic is missing here

x/fibre/blob.go
```diff
@@ -0,0 +1,225 @@
...
+func (d *Blob) RLCOrig() []field.GF128 {
+	return d.rlcOrig
+}
+
+// RowSize returns the size of each row in bytes.
+func (d *Blob) RowSize() int {
+	if len(d.rows) > 0 && len(d.rows[0]) > 0 {
+		return len(d.rows[0])
+	}
+	return 0
```
walldiss: This condition should not be impossible per constructor?

x/fibre/blob.go
```diff
@@ -0,0 +1,225 @@
...
+	if len(d.rows) == 0 {
+		return 0
+	}
+
+	// extract size from header (first row)
+	if len(d.rows[0]) < blobHeaderSize {
+		return 0
+	}
+
+	return int(binary.BigEndian.Uint32(d.rows[0][math.MaxUint32 : math.MaxUint32*2]))
```
walldiss: `[math.MaxUint32 : math.MaxUint32*2]` Would it read from some really high index instead of reading from the 1-5 bytes as encoding spec?

I think unit test would catch it if it used this method

x/fibre/blob.go
```diff
@@ -0,0 +1,225 @@
...
+// The first row is prefixed with a header containing the blob version and original data size,
+// avoiding a full data copy. Returns OriginalRows rows of rowSize bytes each, padding with zeros as needed.
+// Uses slices from the original data when possible, only allocating for the header row and padding.
+func (d *Blob) splitIntoRows(data []byte, rowSize int) [][]byte {
+	rows := make([][]byte, d.cfg.OriginalRows)
+
+	// first row: allocate and write header + beginning of data
+	rows[0] = make([]byte, rowSize)
+	binary.BigEndian.PutUint32(rows[0][0:math.MaxUint32], d.cfg.BlobVersion)
+	binary.BigEndian.PutUint32(rows[0][math.MaxUint32:math.MaxUint32*2], uint32(len(data)))
```
walldiss: I initially thought encoding part was missing. Can we do encoding outside of `split` function? So it would be more of
1. encode with prefixed version, length
2. split

walldiss:
```
binary.BigEndian.PutUint32(rows[0][0:math.MaxUint32], d.cfg.BlobVersion)
binary.BigEndian.PutUint32(rows[0][math.MaxUint32:math.MaxUint32*2], uint32(len(data)))
```

Indexes seems off here. Shoud it be [0:1] for version and [1:5] for length instead?

walldiss (commented):
I think it is worth to add simple encode <-> decode unit test + calling public method to ensure logic is correct.

x/fibre/blob.go
Wondertan: These still gonna be defined by the protocol. The config will be populated by the layer above setting correct protocol constants.

x/fibre/blob.go
Wondertan: The Commitment type is an array which requires slicing to make it a slice.

x/fibre/blob.go
Wondertan: Yes, there is

x/fibre/blob.go
Wondertan: There will be another constructor for the download path which can trigger this.

x/fibre/blob.go
Wondertan: You are right and that's a good catch. I will explain the backstory about what happened here. 

Before putting the `math.MaxUint32` there was `binary.MaxVarintLen32` which claude put there. Before pushing this version I rereviewed the code again and realized that `binary.MaxVarintLen32` is wrong as its actually 5 bytes as opposed to expected 4 and while rushing to push this I put the `math.MaxUint32` which I expected to be the max bytes uint32 takes (4), rather then max value of uint32.  

The other part of the story is that I didn't rerun tests after doing this last minute change and the reason is that this code here is extracted from other little repo that contains all the code and the extraction was without tests as per comment in the PR.

x/fibre/blob.go
Wondertan: The version is uint32 as well in this implementation, the same type as in payment promise proto. Should we make it a single byte here instead?

x/fibre/blob.go
Wondertan: The other problem is that this repo doesn't actually have CI turned on, meaning if there was a test we could still have this slip, if i didn't run it locally.

x/fibre/blob.go
Wondertan: Let me see what I can do here. I worry this is going mess with allocations or may need to leak more fields into the struct just to keep these two funcs decoupled.

walldiss: Yes. 1 byte is enough for version
https://github.com/celestiaorg/fibre-da-spec/blob/main/encoding.md#rsema1d-row-encoding

The spec is incorrectly assumes encoding of every row with version tho. It should be **blob** encoding instead of row row encoding, which you made right in implementation. I'll clarify the encoding spec

Wondertan: And I will update the implemenatation

Wondertan:
I made [da38805](https://github.com/celestiaorg/celestia-app-fibre/pull/14/commits/da388056d41e61d45af58cef09eadf80f3ecea39)

This implements a minor refactor which extracts blob en(de)coding logic into new `blobHeaderV0` struct which encapsulates the pecularities of v0 + I added new tests particularly for the encoding and decoding, so we don't merge any PRs without tests, even if they come later.

x/fibre/blob.go
```diff
@@ -0,0 +1,307 @@
...
+	"errors"
+	"fmt"
+	"runtime"
+
+	"github.com/celestiaorg/rsema1d"
+	"github.com/celestiaorg/rsema1d/field"
+)
+
+// ErrBlobTooLarge is returned when the blob size exceeds MaxBlobSize.
+var ErrBlobTooLarge = errors.New("blob size exceeds maximum allowed size")
```
rootulp, line 15: [nit] can this include Fibre to differentiate it from the existing blobs?

```suggestion
// ErrFibreBlobTooLarge is returned when the Fibre blob size exceeds MaxFibreBlobSize.
var ErrFibreBlobTooLarge = errors.New("fibre blob size exceeds maximum allowed size")
```

x/fibre/blob.go
```diff
@@ -0,0 +1,307 @@
...
+
+// String returns the hex-encoded string representation of the commitment.
+func (c Commitment) String() string {
+	return hex.EncodeToString(c[:])
+}
+
+// BlobConfig contains configuration for erasure coding and data handling.
+type BlobConfig struct {
+	// OriginalRows is the number of original rows before erasure coding.
+	// Here we use explicit naming: OriginalRows (K in rsema1d) and ParityRows (N in rsema1d).
```
rootulp, line 38: 👏 thank you for using explicit naming!

x/fibre/blob.go
```diff
@@ -0,0 +1,307 @@
...
+	// SPECDO: The spec uses N to represent total rows (original + parity), while rsema1d defines N as parity only.
```
rootulp, line 39: [nit] seems like a TODO we can extract to an issue in rsema1d and the specs repo. IMO we shouldn't use N in either place and instead use:

- `OriginalRows`
- `ParityRows`
- `TotalRows`

similar to the code you have here 👍

x/fibre/blob.go
```diff
@@ -0,0 +1,307 @@
...
+	OriginalRows int
+	// ParityRows is the number of parity rows added by erasure coding.
+	// Total rows = OriginalRows + ParityRows.
+	ParityRows int
+	// RowSizeMin is the minimum row size in bytes.
+	RowSizeMin int
+	// MaxBlobSize is the maximum allowed blob size.
+	MaxBlobSize int
+	// BlobVersion is the version of the row format.
+	BlobVersion uint8
```
rootulp, line 49: [uber nit] to clarify these are for Fibre blobs. I guess it's somewhat implied by this living in the x/fibre package but we already have constants in celestia-app for the existing blobs so a Fibre in the name would help differentiate.


```suggestion
	// MaxFibreBlobSize is the maximum allowed Fibre blob size.
	MaxFibreBlobSize int
	// FibreBlobVersion is the version of the row format.
	FibreBlobVersion uint8
```

x/fibre/blob.go
```diff
@@ -0,0 +1,307 @@
...
+	CodingWorkers int
+}
+
+// DefaultBlobConfig returns a [BlobConfig] with default values.
+func DefaultBlobConfig() BlobConfig {
+	return BlobConfig{
+		OriginalRows:  1 << 12,   // 4096
+		ParityRows:    3 << 12,   // 12288 (3 * 4096), total rows = 16384
+		RowSizeMin:    1 << 6,    // 64 bytes
+		MaxBlobSize:   128 << 20, // 128 mib
```
rootulp: the comments aren't necessary if we update the constants


```suggestion
		OriginalRows:  4096,
		ParityRows:    12288,
		RowSizeMin:    64,
		MaxBlobSize:   128 * mebibyte,
```

maybe I'm missing something on why these constants use bitwise operations. Is this a type of performance optimization?

rootulp (commented):
LGTM but CI is failing, potentially because the rsema1d import

x/fibre/blob.go
Wondertan: The actual name of this var is `fibre.ErrBlobTooLarge`. Basically everything in this pkg is already domain separated, so I don't see much reason to duplicate the separator

x/fibre/blob.go
Wondertan: >  I guess it's somewhat implied by this living in the x/fibre package but we already have constants in celestia-app for the existing blobs so a Fibre in the name would help differentiate.

If we do it for blob then we should also do it for all the mentions of Row, like RowSize or MaxRowSize, etc, because row do also exist in the core protocol. I believe the `fibre` as pkg is a sufficient domain separator and adding it again within the code for everything that might be similar with core protocol is superfluous.

x/fibre/blob.go
Wondertan: I don't mind renaming it everywhere else. Descriptiveness rarely hurts anybody before it becomes verbosity.

x/fibre/blob.go
Wondertan: It just looks cooler 🕶️ 

and I am trying to remember which power of 2 is which number

x/fibre/blob.go
rootulp: Fair enough, the type is in the `fibre` package so it doesn't need it. 

The error message may still warrant a "fibre" because we already have these:

https://github.com/celestiaorg/celestia-app/blob/e93e3c24e729e734381266b15fd36b87cfd48352/x/blob/types/errors.go#L37-L38

but I'll defer to you.

Wondertan: agree on the error string

x/fibre/blob.go
rootulp: I don't think RowSize or MaxRowSize need a Fibre prefix because celestia-app doesn't have constants for those already so there isn't a real risk of mixing those up with existing terms in celestia-app.

On the other hand, there are existing blobs in celestia-app and in celestia-app v3.x there was even an ante handler [here](https://github.com/celestiaorg/celestia-app/blob/4dd92890ac0a50d163627053d6e0323c9b5ba6a9/x/blob/ante/max_total_blob_size_ante.go#L46-L57) which is really similar to the `MaxBlobSize` idea expressed here.

Wondertan: > I don't think RowSize or MaxRowSize need a Fibre prefix because celestia-app doesn't have constants for those already so there isn't a real risk of mixing those up with existing terms in celestia-app.

So why wouldn't we add the prefix to the constants we put them together with constants for core protocol? This is just a config that is agnostic to what protocol is and doesn't know if there is some other core params with similar names. For this config, prefixing Fibre is redundant. For the place which instaniates the config - it will be not, so we can keep it there.

rootulp: Resolving b/c I'm not blocking on adding the Fibre in the error message or name, it was just a suggestion.

x/fibre/blob.go
rootulp: I assume that's sarcasm so let's update to my suggestion and resolve this thread.

Wondertan: Can't apply the suggestion as there is no mebibyte literal, but updated manually

x/fibre/blob.go
Wondertan: Updated to be a single byte

Wondertan: Should be better now

x/fibre/blob.go
Wondertan: This fixed now by defining consts manually

x/fibre/blob.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+const (
+	// uint32SizeBytes is the size of a uint32 in bytes.
+	uint32SizeBytes = 4
+	// uint8SizeBytes is the size of a uint8 in bytes.
+	uint8SizeBytes = 1
+	// blobHeaderSize is the size of the blob header in bytes.
+	// Format: 1 byte version (uint8) + 4 bytes blob size (uint32)
+	blobHeaderSize = uint8SizeBytes + uint32SizeBytes
```
walldiss: Was this meant to be encoding constants? Those should probably reflect it by being named not for uint length, but for encoding meaning:
```suggestion
	// blobLengthSizeBytes is the size of a blob length encoding in bytes.
	blobLengthSizeBytes = 4
	// blobVersionSizeBytes is the size of a blob version encoding in bytes.
	blobVersionSizeBytes = 1
	// blobHeaderSize is the size of the blob header in bytes.
	// Format: 1 byte version (uint8) + 4 bytes blob size (uint32)
	blobHeaderSize = blobVersionSizeBytes + dataLengthSizeBytes
```

x/fibre/blob.go
```diff
@@ -0,0 +1,307 @@
...
+
+	// decode header from first row
+	if err := h.decode(rows[0]); err != nil {
+		return nil, fmt.Errorf("decoding header: %w", err)
+	}
+
+	size := blobHeaderSize + int(h.dataSize)
+
+	// Pre-allocate the exact size needed
+	data := make([]byte, size)
```
walldiss: Please validate 
1. that blob_size parsed from header is within reasonable bounds. 
2. bytes in array match the blob_size parsed

Later check for `copied < size` would not be needed.

x/fibre/blob.go
```diff
@@ -0,0 +1,307 @@
...
+	// Return original data without header
+	return data[blobHeaderSize:size:size], nil
+}
+
+// calculateRowSize computes the row size for the given data length and config.
+// Row size is calculated as ceil((dataLen + headerSize) / OriginalRows),
+// rounded up to the nearest multiple of RowSizeMin.
+func (h blobHeaderV0) calculateRowSize(dataLen int, cfg BlobConfig) int {
+	totalLen := dataLen + blobHeaderSize
+	minRowSize := (totalLen + cfg.OriginalRows - 1) / cfg.OriginalRows // ceil(totalLen / OriginalRows)
```
walldiss, line 279: why not just math.Ceil? feels harder to read as is.

x/fibre/blob.go
Wondertan: Agree, this can be better. 

I am going for 

```
const (
	// blobVersionLen is the length of the version field in bytes.
	blobVersionLen = 1
	// blobDataSizeLen is the length of the data size field in bytes.
	blobDataSizeLen = 4
	// blobHeaderLen is the total length of the blob header in bytes.
	// Format: 1 byte version + 4 bytes data size
	blobHeaderLen = blobVersionLen + blobDataSizeLen
)
```

Using Len is more idiomatic

x/fibre/blob.go
Wondertan: I added a change in the client, which prefixes any returned error with "fibre: ", even if it's not predefined one

walldiss (approved):

PR #24, opened 2025-10-16
feat: add validator address module x/valaddr

mcrakhman:
## Overview

Added the `x/valaddr` module according to the spec https://github.com/celestiaorg/fibre-da-spec/pull/4.


x/valaddr/keeper/msg_server.go
```diff
@@ -0,0 +1,67 @@
...
+		return nil, errors.Wrapf(types.ErrIncorrectValidator, "failed to get consensus public key: %v", err)
+	}
+	consAddr := sdk.ConsAddress(consPubKey.Address())
+
+	if len(msg.IpAddress) > types.MaxIPLen {
+		return nil, errors.Wrapf(types.ErrInvalidIPAddress, "IP address must be less than 90 characters, got %d", len(msg.IpAddress))
+	}
+	if net.ParseIP(msg.IpAddress) == nil {
+		return nil, errors.Wrapf(types.ErrInvalidIPAddress, "invalid IP address: %s", msg.IpAddress)
+	}
```
evan-forbes: might be wrong but I think we ended up not wanting this as validators are almost expected to use some dns

if the spec still says this we should also edit it

basically this just needs to be a string that's bounded in length

proto/celestia/valaddr/v1/params.proto
```diff
@@ -0,0 +1,11 @@
...
+package celestia.valaddr.v1;
+
+option go_package = "github.com/celestiaorg/celestia-app/x/valaddr/types";
+
+// Params defines the parameters for the valaddr module.
+message Params {
+  // missing_info_check_height is the height at which to check for missing validator info
+  // after the upgrade (if upgrade was at height X, then we check at X + missing_info_check_height)
+  int64 missing_info_check_height = 1;
+}
```
rootulp: [question] I don't understand why this is a fixed height instead of a periodic interval. I assume the state machine checks for missing info on a periodic basis rather than just once 100,000 blocks after v7 activates.

Relevant: https://github.com/celestiaorg/fibre-da-spec/pull/4#discussion_r2460847688

proto/celestia/valaddr/v1/query.proto
```diff
@@ -0,0 +1,57 @@
...
+// Query defines the gRPC query service for the valaddr module.
+service Query {
+  // FibreProviderInfo queries the fibre provider information for a specific validator
+  rpc FibreProviderInfo(QueryFibreProviderInfoRequest) returns (QueryFibreProviderInfoResponse) {
+    option (google.api.http).get = "/valaddr/v1/fibre-provider-info/{validator_consensus_address}";
+  }
+
+  // AllActiveFibreProviders queries fibre provider information for all validators in the active set
+  rpc AllActiveFibreProviders(QueryAllActiveFibreProvidersRequest) returns (QueryAllActiveFibreProvidersResponse) {
+    option (google.api.http).get = "/valaddr/v1/active-fibre-providers";
```
rootulp: [nit] idk if we do this elsewhere but do we want the full method name in the path?


```suggestion
    option (google.api.http).get = "/valaddr/v1/all-active-fibre-providers";
```

x/valaddr/client/cli/tx.go
```diff
@@ -0,0 +1,67 @@
...
+		CmdSetFibreProviderIP(),
+	)
+
+	return cmd
+}
+
+// CmdSetFibreProviderIP broadcasts a MsgSetFibreProviderInfo transaction
+func CmdSetFibreProviderIP() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "set-provider-ip [ip-address]",
```
rootulp: [nit] currently `MsgSetFibreProviderInfo` only contains the IP address but I think it may future-proof if this command was called


```suggestion
// CmdSetFibreProviderInfo broadcasts a MsgSetFibreProviderInfo transaction
func CmdSetFibreProviderInfo() *cobra.Command {
	cmd := &cobra.Command{
		Use:   "set-provider-info [ip-address]",
```

that way we can add more fields to ProviderInfo at a later date and not break the name of this command.

Note this suggestion doesn't fix all the instances in this file

x/valaddr/keeper/grpc_query.go
```diff
@@ -0,0 +1,66 @@
...
+
+		providers = append(providers, types.FibreProvider{
+			ValidatorConsensusAddress: consAddr.String(),
+			Info:                      info,
+		})
+	}
+
+	return &types.QueryAllActiveFibreProvidersResponse{
+		Providers: providers,
+	}, nil
```
rootulp: [question] this implementation looks like it gets all bonded validators which I assume is a superset of the active validators.

Is that right? Is there a unit test that confirms this only returns the _active_ validators and not all bonded validators?

x/valaddr/types/errors.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+import (
+	"cosmossdk.io/errors"
+)
+
+var (
+	ErrInvalidIPAddress   = errors.Register(ModuleName, 1, "invalid IP address format")
+	ErrIncorrectValidator = errors.Register(ModuleName, 2, "invalid signer address")
```
rootulp: [nit] the error message and the error name don't match. Can the error or error message be renamed to include more info? For example if this is only returned when a validator does not exist for the provided signer address then perhaps


```suggestion
	ErrInvalidValidator = errors.Register(ModuleName, 2, "invalid validator")
```

and then the usage can be like

```go
return errors.Wrapf(types.ErrInvalidValidator, "signer %v does not match a validator", signer)
```

x/valaddr/types/msg.go
```diff
@@ -0,0 +1,45 @@
...
+	// Validate validator address
+	if m.Signer == "" {
+		return errorsmod.Wrap(ErrIncorrectValidator, "validator address cannot be empty")
+	}
+	_, err := sdk.ValAddressFromBech32(m.Signer)
+	if err != nil {
+		return errorsmod.Wrapf(ErrIncorrectValidator, "invalid validator address: %v", err)
+	}
+
+	// this was in the spec, though it is not necessary because net.ParseIP uses RFC 4291 and it will reject it anyway
```
rootulp: given this comment, should we update the spec and the implementation to remove this check and `MaxIPLen`?

rootulp (dismissed):
Overall LGTM, great work. None of my feedback is blocking so can be addressed in follow-ups.

Given this implements https://github.com/celestiaorg/fibre-da-spec/pull/4, should we try to merge that spec PR with or as a prerequisite to this PR?

proto/celestia/valaddr/v1/params.proto
mcrakhman: Agree, periodic checks makes more sense, because validators may come and go etc. Can we add this as a follow up, when we implement jaling, because we don't use this parameter right now?

x/valaddr/keeper/grpc_query.go
mcrakhman: It uses this function which gets 100 (MaxValidators) validators with most power, so it seems the should be in active set 

```// GetBondedValidatorsByPower gets the current group of bonded validators sorted by power-rank
func (k Keeper) GetBondedValidatorsByPower(ctx context.Context) ([]types.Validator, error) {
```

x/valaddr/types/msg.go
mcrakhman: Use host instead of IP now

app/test/std_sdk_test.go
```diff
@@ -335,12 +335,12 @@ func (s *StandardSDKIntegrationTestSuite) TestStandardSDK() {
 			expectedCode: abci.CodeTypeOK,
 		},
 		{
-			name: "set fibre provider IP address",
+			name: "set fibre provider dns",
```
rootulp:
```suggestion
			name: "set fibre provider host",
```

proto/celestia/valaddr/v1/params.proto
rootulp: Yes can be a follow-up.

x/valaddr/keeper/grpc_query.go
rootulp: oh good to know, thanks for investigating!

x/valaddr/types/errors.go
```diff
@@ -5,6 +5,6 @@ import (
 )
 
 var (
-	ErrInvalidIPAddress   = errors.Register(ModuleName, 1, "invalid IP address format")
-	ErrIncorrectValidator = errors.Register(ModuleName, 2, "invalid signer address")
+	ErrInvalidHostAddress = errors.Register(ModuleName, 1, "invalid address")
```
rootulp:
```suggestion
	ErrInvalidHostAddress = errors.Register(ModuleName, 1, "invalid host address")
```

rootulp (dismissed):
LGTM

x/valaddr/keeper/msg_server.go
cmwaters: Yeah I've update the spec. It should be a string no longer than 90 in length

proto/celestia/valaddr/v1/params.proto
cmwaters: I can explain the reasoning. It's a static height because the idea was that we only need to check this once (some grace period after migrating to v7) just to assess the current validator set. All new validators that enter the set have another check that will jail them if they don't have the fibre server address set.

x/valaddr/client/cli/tx.go
```diff
@@ -0,0 +1,67 @@
...
+		CmdSetFibreProviderInfo(),
+	)
+
+	return cmd
+}
+
+// CmdSetFibreProviderInfo broadcasts a MsgSetFibreProviderInfo transaction
+func CmdSetFibreProviderInfo() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "set-provider-info [host]",
```
cmwaters:
```suggestion
		Use:   "set-host [host]",
```
This is what I've used in the spec to keep terminology consistent

x/valaddr/types/msg.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+const (
+	// EventTypeSetFibreProviderInfo is the event type for setting fibre provider info
+	EventTypeSetFibreProviderInfo = "set_fibre_provider_info"
+
+	// AttributeKeyValidatorAddress is the attribute key for consensus address
+	AttributeKeyValidatorAddress = "validator_consensus_address"
+	// AttributeKeyIPAddress is the attribute key for IP address
+	AttributeKeyIPAddress = "ip_address"
```
cmwaters: Shall we be using host to be consistent

app/app.go
```diff
@@ -183,6 +186,7 @@ type App struct {
 	CircuitKeeper       circuitkeeper.Keeper
 	HyperlaneKeeper     hyperlanekeeper.Keeper
 	WarpKeeper          warpkeeper.Keeper
+	ValaddrKeeper       valaddrkeeper.Keeper
```
cmwaters: nit: but looks more readable this way
```suggestion
	ValAddrKeeper       valaddrkeeper.Keeper
```

cmwaters (commented):
Good work here!

Can we add an integration test, where users are submitting a transaction to set the info using TxClient and then querying it. Specifically, I want to check that a end user can query the validator set and query the fibre providers and deduce which providers are in the active set

Secondly, this doesn't implement any of the jailing logic which I'm fine with for now but for atomicity, I would prefer if the params which only relate to jailing logic weren't included here and were simply added later

proto/celestia/valaddr/v1/genesis.proto
```diff
@@ -1,13 +1,7 @@
...
-import "celestia/valaddr/v1/params.proto";
-
 option go_package = "github.com/celestiaorg/celestia-app/x/valaddr/types";
 
 // GenesisState defines the valaddr module's genesis state.
-message GenesisState {
-  // params defines the module parameters
-  Params params = 1 [(gogoproto.nullable) = false];
-}
+message GenesisState {}
```
rootulp, line 7: [question] why are there no more params? Maybe params will be added in a follow-up PR

rootulp (approved):

proto/celestia/valaddr/v1/genesis.proto
cmwaters: There are no need for params until we have the jailing mechanism in place. Because this PR does not introduce the jailing mechanism, it's confusing to add Params that aren't used and more fitting to add it later

cmwaters (approved):
Good clean implementation of the module

proto/celestia/valaddr/v1/genesis.proto
rootulp: Got it, makes sense

PR #25, opened 2025-10-17
feat(fibre): `Client` with `Upload` and `Put`

Wondertan:
Finally, the `Client` with the first two methods: `Upload` and `Put`. `Put` follows the spec with the input and output params. The `Upload`, on the other hand, is a lower-level method that gives the user more control and is used by `Put`. Besides, this PR also moves what it can out of `x/fibre` to `fibre`.

The other notable aspect of this PR is the benchmarks. They demonstrate the throughput of `Put/Upload` for low, middle, and high blob sizes, and with various parallel upload params. It benchmarks the entire process from blob encoding to signature verification, with the server and networking functionality entirely mocked or cached out. The results are well documented next to the [benchmarks](https://github.com/celestiaorg/celestia-app-fibre/pull/25/files#diff-804775a9c7679c01b4e16949ddce73d00cf3906401e857b421526d9d87a6447bR86-R109). Besides showing how well concurrent uploads scale, they also hint us that the next bottleneck will be either on grpc, db or even in state machine handling PPs. 

The client was written with great care and attention to detail. The code is simplified as much as possible, and if you see anything else that can be simplified, please let me know. The trace coverage is complete, aiding future bottleneck discovery. The test cases also cover the most critical security cases, but I could miss some, and I ask you to brainstorm some more while reviewing.

The PR was made bigger as requested, but I am still keeping the download portion for the following PR. 

All the tests pass with race detector and `--count 100`, so they are not flaky, and golanci-lint succeeds.


fibre/client.go
```diff
@@ -0,0 +1,130 @@
...
+	cfg ClientConfig
+
+	txClient *user.TxClient
+	keyring  keyring.Keyring
+	valGet   validator.SetGetter
+	hostReg  validator.HostRegistry
+
+	log    *slog.Logger
+	tracer trace.Tracer
+	clock  clock.Clock
```
walldiss, line 88: Just curious what is the reason to use custom clock here instead of time pkg?

fibre/client.go
```diff
@@ -0,0 +1,130 @@
...
+func (c *Client) Config() ClientConfig {
+	return c.cfg
+}
+
+// Close closes the client and releases any associated resources.
+// It waits for all ongoing Put/Get operations to complete before closing.
+func (c *Client) Close() error {
+	c.closeWg.Wait()
+	return c.clientCache.Close()
+}
```
walldiss, line 155: If we want graceful shutdown Close() should also prevent allowing of new Puts/Gets. Current logic does not guarantee that, because new woperation can be created after wg is done and before clientCache is closed. non-blocking: Can be solved as followup PR

fibre/client_put.go
```diff
@@ -0,0 +1,74 @@
...
+	))
+
+	signedPromise, err := c.Upload(ctx, ns, blob)
+	if err != nil {
+		span.RecordError(err)
+		span.SetStatus(codes.Error, "failed to upload blob")
+		return result, err
+	}
+
+	// TODO(@Wondertan): Submit and await inclusion of MsgPayForFibre transaction through txClient.
```
walldiss, line 66: This is major step. Is this call decoupled to followup PR or are there any blockers to implement it? Don't see anything about it being mentioned in PR description

fibre/client_upload.go
```diff
@@ -0,0 +1,301 @@
...
+// Upload uploads the given [Blob] to the Fibre network.
+// It creates a [PaymentPromise], uploads the data to validators, and collects signatures confirming the upload.
+// Returns a [SignedPaymentPromise] containing the promise and validator signatures.
+// May keep uploading data in background after returning successfully.
+func (c *Client) Upload(ctx context.Context, ns share.Namespace, blob *Blob) (result SignedPaymentPromise, err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("fibre upload: %w", err)
+		}
+	}()
```
walldiss: Kinda nit, but I would like to advocate to not do this defer because it is a footgun. I've seen legs lost on similar code in the past. If variable is shadowed it will miss  wrap on return.

fibre/client.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	if cfg.Tracer == nil {
+		cfg.Tracer = otel.Tracer("fibre")
+	}
+	if cfg.Log == nil {
+		cfg.Log = slog.Default()
+	}
+	if cfg.Clock == nil {
+		cfg.Clock = clock.New()
+	}
```
walldiss: check that default key exists

fibre/client_upload.go
```diff
@@ -0,0 +1,323 @@
...
+	if err != nil {
+		c.log.WarnContext(ctx, "failed to get client for validator", "validator", val.Address.String(), "error", err)
+		span.RecordError(err)
+		span.SetStatus(codes.Error, "failed to get client")
+		return
+	}
+	span.AddEvent("client_acquired")
+
+	// generate proofs for assigned rows in
+	// we do this in upload routines to parallelize proofs generation which show 50-64% performance increase for 256KB-4MB blobs
```
walldiss: We should generate merkle tree once. And then construct proofs from it. Should be good for client CPU. (non-blocking)

fibre/client_upload.go
```diff
@@ -0,0 +1,323 @@
...
+		responsesExhaustedCh = make(chan struct{}) // closes when all responses complete
+	)
+
+	for val, req := range requests {
+		// acquire semaphore before spawning goroutine
+		select {
+		case c.uploadSem <- struct{}{}:
+		case <-ctx.Done():
+			return ctx.Err()
+		}
```
walldiss, line 260: I was thinking of worker pool per upload, but it is global here. Don't think it was explicitly stating the which one should be in spec. Wonder if you compared both and picked global one and if yes what were pros/cons

fibre/client_upload.go
```diff
@@ -0,0 +1,323 @@
...
+		c.closeWg.Add(1)
+		go func(val *core.Validator, req *types.UploadRowsRequest) {
+			defer func() {
+				// release semaphore
+				<-c.uploadSem
+
+				// increment responses and mark as completed if so
+				if int(responses.Add(1)) == len(requests) {
+					close(responsesExhaustedCh)
+				}
```
walldiss, line 271: What is the case that you try to cover with counting responses? I would guess it is counting failed responses that do not result in signature?

fibre/client_upload.go
```diff
@@ -0,0 +1,323 @@
...
+			Promise: pbPromise,
+			Rows:    make([]*types.Row, 0, len(rowIndices)),
+			RlcOrig: rlcOrigBytes,
+		}
+		for _, rowIndex := range rowIndices {
+			req.Rows = append(req.Rows, &types.Row{
+				Index: uint32(rowIndex),
+			})
+		}
+		requests[val] = req
```
walldiss, line 315: Does it fill only index field? Seems like rows data is missing here?

fibre/client_upload_bench_test.go
```diff
@@ -0,0 +1,247 @@
...
+//
+// CPU: AMD Ryzen 9 7940HS w/ Radeon 780M Graphics
+// Results with 100 validators (averaged over 5 iterations):
+//
+//	Blob Size      Time/op    Throughput    Memory/op    Allocs/op
+//	1 KiB          ~20.7 ms   ~0.047 MiB/s  ~29.3 MB     ~253k
+//	16 KiB         ~20.8 ms   ~0.75 MiB/s   ~29.3 MB     ~252.5k
+//	128 KiB        ~20.8 ms   ~6.0 MiB/s    ~29.2 MB     ~251k
+//	1 MiB          ~23.5 ms   ~42.6 MiB/s   ~41.9 MB     ~250k
+//	128 MiB (max)  ~725 ms    ~177 MiB/s    ~1,776 MB    ~294k
```
walldiss: Everything <1Mib is just smallest fibre blob and overhead is the same. Would be nice to have like 1,4,16,64 Mib instead

fibre/client_upload_bench_test.go
```diff
@@ -0,0 +1,247 @@
...
+//	Blob Size    Concurrency    Time/op     Throughput     Memory/op    Allocs/op
+//	128 KiB      1              ~20.4 ms    ~6.1 MiB/s     ~29.2 MB     ~251k
+//	128 KiB      4              ~27.9 ms    ~17.9 MiB/s    ~114 MB      ~986k
+//	128 KiB      8              ~42.5 ms    ~23.5 MiB/s    ~223 MB      ~1.94M
+//	1 MiB        1              ~23.5 ms    ~42.6 MiB/s    ~41.9 MB     ~250k
+//	1 MiB        4              ~42.1 ms    ~95.1 MiB/s    ~165.7 MB    ~987k
+//	1 MiB        8              ~80.1 ms    ~99.9 MiB/s    ~329 MB      ~1.96M
+//	1 MiB        16             ~168 ms     ~95.4 MiB/s    ~647 MB      ~3.85M
+//	128 MiB      1              ~716 ms     ~179 MiB/s     ~1,776 MB    ~294k
+//	128 MiB      4              ~2,791 ms   ~184 MiB/s     ~7,103 MB    ~1.18M
```
walldiss: Nice! It proofs additional external concurrency does not improve throughput. So the code is well optimised for concurrency under the hood.

walldiss (dismissed):
Massive work and good quality! Few questions here and there, but generally looks good

fibre/client_upload.go
Wondertan: This is done in `uploadToValidator`. Requests are prepared here and then populated with rows and proofs in validator-specific routines to parallelise, as it brings up to 60% of performance.

fibre/client_upload.go
Wondertan: If we count that success is handed by the `Done`, then yes, only for failures, so we return to the client without needing them to wait.

fibre/client_upload.go
Wondertan: Otherwise, the number of uploads is unbounded, breaking the original intention of controlling the resources.

fibre/client_upload.go
Wondertan: This is where we set the rows https://github.com/celestiaorg/celestia-app-fibre/pull/25#discussion_r2444764674

fibre/client.go
Wondertan: valid

fibre/client_upload.go
Wondertan: The goal here is to make sure all the errors returned from the client are domain separated with fibre to not be mixed with regular blobs/rows/etc. While I agree this might be a footgun, but if it's done by the hands who know what they are doing, then it's fine. But I am also happy to hear the alternative options if any

fibre/client_put.go
Wondertan: The state logic is simply not implemented.

fibre/client.go
Wondertan: To mock it out in the benchmarks. The std time does not allow you to do this unfortunately.

fibre/client.go
Wondertan: Yes, we could return an error saying the client is closed

fibre/client_upload_bench_test.go
Wondertan: > It proofs additional external concurrency does not improve throughput.

It does for 1MiB blobs

fibre/client_upload.go
walldiss: It can be solved by caller. Caller knows it is doing fibre operations, so will not be confused. The wrapping can happen outside the Upload function:

err = c.Upload
if err != nil {
      return fmt.Errorf("fibre upload: %w")
}


Imo we can just remove this defer wrapping.

fibre/client_upload.go
Wondertan: On the comment. The tree is already generated there, and we are simply parallelizing allocations of the resulting proof.

walldiss: I think optimisation with one run tree calculation still stands? It would be similar to proofs caching

fibre/client.go
Wondertan: > // Sign the real PaymentPromise data on first call and cache it.
	// We must sign the actual PaymentPromise because SignatureSet verifies signatures against
	// the real sign bytes. Since we use a mocked clock, all PaymentPromises have the same
	// CreationTimestamp, making their SignBytes identical, so we can safely cache the signature.
	// sync.Once ensures lock-free reads after the first initialization.

fibre/client_upload.go
Wondertan: The tree is already computed once when we Extend. The comment might be confusing there by saying "proof generation", I will update

fibre/client_upload.go
Wondertan: Agreed that we can defer this to user

fibre/client_upload_bench_test.go
Wondertan: The other thing to note is when you run the benchmark single time. I can get this (--count 1 vs count 5)
> BenchmarkClient_Upload_Concurrent/128_MiB_max/concurrency_4-16     	       1	1216153055 ns/op	       **421.0 MiB/s**	1806953032 B/op	  592583 allocs/op

walldiss (dismissed):

walldiss (dismissed):

fibre/client_upload_bench_test.go
Wondertan: Somehow, I got with count 5
> // 128 MiB 4 ~777 ms ~660 MiB/s ~1,807 MB ~593.3k

walldiss (approved):

walldiss (approved):

walldiss (approved):

PR #26, opened 2025-10-19
chore: private repo ci

evan-forbes:
~~still a wip, trying to fix ci for private repos~~

note that we will ofc want to revert this before merging upstream


evan-forbes (commented):
this is 110% AI slop and I don't fully understand each change, and there might be better ways to add the secret instead of modifying each run, however it does work. 

We will simply revert before merging so its very temporary tech debt :shrug: #36

.github/workflows/test.yml
```diff
@@ -1,20 +1,42 @@
...
+          restore-keys: |
+            ${{ runner.os }}-go-${{ steps.go.outputs.go-version }}-
+
+      - name: Download Go modules
+        run: go mod download
 
       - name: Run tests in short mode
         run: make test-short
-        timeout-minutes: 10
+        timeout-minutes: 25
```
rootulp, line 39: seems like test run time for test-short should not have changed so proposal to revert or skip any tests that take longer than 10 minutes from test-short


```suggestion
        timeout-minutes: 10
```

docker/multiplexer-local.Dockerfile
```diff
@@ -0,0 +1,132 @@
+# This Dockerfile performs a multi-stage build. BUILDER_IMAGE is the image used
```
rootulp, line 1: [question] why do we need this? I couldn't find a CI workflow that used it.

Makefile
```diff
@@ -202,6 +202,23 @@ build-docker-multiplexer:
...
+		exit 1; \
+	fi
+	@export DOCKER_BUILDKIT=0 && \
+	$(DOCKER) build \
+		--build-arg GITHUB_TOKEN="$$GH_PRIVATE_REPO" \
+		--build-arg TARGETOS=$(DOCKER_GOOS) \
+		--build-arg TARGETARCH=$(DOCKER_GOARCH) \
+		-t ghcr.io/celestiaorg/celestia-app:$(CELESTIA_TAG) \
+		-f docker/multiplexer-local.Dockerfile .
+.PHONY: build-docker-multiplexer-local
```
rootulp, line 220: [question] why do we need this? I couldn't find a CI workflow that invoked it.

Makefile
```diff
@@ -315,7 +332,7 @@ test-race:
...
+	@go test -timeout 15m -v -race -skip "TestPrepareProposalConsistency|TestIntegrationTestSuite|TestSquareSizeIntegrationTest|TestStandardSDKIntegrationTestSuite|TestTxsimCommandFlags|TestTxsimCommandEnvVar|TestTxsimDefaultKeypath|TestMintIntegrationTestSuite|TestUpgrade|TestMaliciousTestNode|TestBigBlobSuite|TestQGBIntegrationSuite|TestSignerTestSuite|TestPriorityTestSuite|TestTimeInPrepareProposalContext|TestCLITestSuite|TestLegacyUpgrade|TestSignerTwins|TestConcurrentTxSubmission|TestTxClientTestSuite|Test_testnode|TestEvictions|TestEstimateGasUsed|TestEstimateGasPrice|TestWithEstimatorService|TestTxsOverMaxTxSizeGetRejected|TestStart_Success|TestReadBlockchainHeaders|TestPrepareProposalCappingNumberOfMessages|TestGasEstimatorE2E|TestGasEstimatorE2EWithNetworkMinGasPrice|TestRejections|TestClaimRewardsAfterFullUndelegation|TestParallelTxSubmission|TestRun" ./...
```
rootulp, line 335: [question] why add the `./...` at the end?

rootulp (approved):
A few non-blocking questions. LGTM if CI passes but I think we may be able to revert some changes in this PR that aren't strictly necessary to make CI pass

Makefile
rootulp: Oh nvm, just realized there was a `./...` at the beginning. AI just moved it

.github/workflows/test.yml
evan-forbes: yeah there's something weird with the private repo thing that this has to redownload all the go mods again and it can take a while

not sure why we're seeing this tbh

Makefile
evan-forbes: moving it was an attempt to get the chainbuilder test to get ignored as for some reason it doesn't get ignored despite us adding it to the list of tests to ignore during the make test-race

Makefile
evan-forbes: not sure maybe we can delete

evan-forbes:
I'm going to squash and merge to progress with the rebase thing, but will dig into reverting the multiplexer build before we remove everything in #36

PR #29, opened 2025-10-21
fix(proto/fibre): send rlc root once

Wondertan:
The previous version of proto sent RLC root per every row, while there is only one rlc root for the entire blob.

Besides, the Rows message is introduced with the `oneof` construct. That is, the rows can be sent with either the RLC root or the RLC coefficients of the original rows. The Rows are used now for uploads, downloads, and serialization on the server's storage.

Also, the RLC coefficients are now sent as a single array, as we always know its size and do not need protobuf to prefix varint to it.

### DIFF
--- a/fibre/client_upload.go
+++ b/fibre/client_upload.go
@@ -77,7 +77,7 @@ func (c *Client) Upload(ctx context.Context, ns share.Namespace, blob *Blob) (re
 		return result, fmt.Errorf("preparing bytes to sign: %w", err)
 	}
 
-	requests := makeUploadRequests(shardMap, promise.ToProto(), blob.RLCOrig())
+	requests := makeUploadRequests(shardMap, promise.ToProto(), blob.RLCCoeffs())
 	sigSet := valSet.NewSignatureSet(c.cfg.UploadTargetVotingPower, c.cfg.UploadTargetSignaturesCount, signBytes)
 
 	c.log.DebugContext(ctx, "initiating blob upload",
@@ -181,7 +181,7 @@ func (c *Client) uploadToValidator(
 	ctx, span := c.tracer.Start(ctx, "fibre.upload_to_validator",
 		trace.WithAttributes(
 			attribute.String("validator_address", val.Address.String()),
-			attribute.Int("row_count", len(req.Rows)),
+			attribute.Int("row_count", len(req.Rows.Rows)),
 		),
 	)
 	defer span.End()
@@ -197,7 +197,7 @@ func (c *Client) uploadToValidator(
 	span.AddEvent("client_acquired")
 
 	// get proofs and rows here in per request routine which is in parallel which ~39% faster for max blob size
-	for i, rowPb := range req.Rows {
+	for i, rowPb := range req.Rows.Rows {
 		row, err := blob.Row(int(rowPb.Index))
 		if err != nil {
 			c.log.WarnContext(ctx, "failed to generate proof for row", "validator", val.Address.String(), "row_index", rowPb.Index, "error", err)
@@ -205,8 +205,8 @@ func (c *Client) uploadToValidator(
 			span.SetStatus(codes.Error, "failed to generate proof")
 			return
 		}
-		req.Rows[i].Data = row.Row
-		req.Rows[i].Proof = row.RowProof
+		req.Rows.Rows[i].Data = row.Row
+		req.Rows.Rows[i].Proof = row.RowProof
 	}
 	span.AddEvent("proofs_generated")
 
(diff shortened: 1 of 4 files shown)


proto/celestia/fibre/v1/service.proto
```diff
@@ -5,29 +5,44 @@ option go_package = "github.com/celestiaorg/celestia-app/x/fibre/types";
...
 }
 
+// Rows represents a collection of rows in a Fibre blob.
+// With two ways of proving data correctness:
+// 1. Using RLC coefficients when there is not enough rows to reconstruct the data blob. (Upload case)
+// 2. Using a Merkle root of RLC tree, when there is sufficient rows to reconstruct the data blob. (Download case)
+message Rows {
+  repeated Row rows = 1;
+  oneof rlc {
+    bytes coefficients = 2; // serialized rlc coefficients, 16 bytes per original row
```
rootulp: [optional] since this isn't a slice of slices, it looks like all 16 byte coefficients get flattened into one slice.

```suggestion
    bytes coefficients = 2; // serialized flattened rlc coefficients, 16 bytes per original row
```

fibre/client_upload.go
```diff
@@ -181,7 +181,7 @@ func (c *Client) uploadToValidator(
 	ctx, span := c.tracer.Start(ctx, "fibre.upload_to_validator",
 		trace.WithAttributes(
 			attribute.String("validator_address", val.Address.String()),
-			attribute.Int("row_count", len(req.Rows)),
+			attribute.Int("row_count", len(req.Rows.Rows)),
```
rootulp, line 184: [optional]

`Rows.Rows` seems a bit strange. Maybe we can differentiate the first `Rows` because it also contains RLC metadata. A few candidates but open to alternatives:

1. `RowsWithRlc.Rows`
2. `RowsWithMetadata.Rows`

rootulp (dismissed):

proto/celestia/fibre/v1/service.proto
```diff
@@ -5,29 +5,44 @@ option go_package = "github.com/celestiaorg/celestia-app/x/fibre/types";
...
+    bytes root = 3; // const len == 32 when present
```
Wondertan, line 23: Another opinion is that we may not need the root at all, since the client can recompute it as well. However, then we lose the ability to verify the response from the validator independently of other validators. Currently, when the client receives a portion of rows from a validator, it can immediately determine whether the rows are included in the commitment, and only after it has the necessary data, it verifies the RLC. The alternative is to not care about including independent validator responses and to confirm everything at once in the end. Considering the security model assumes an honest majority anyway, removing the root sounds like a valid thing to do.

proto/celestia/fibre/v1/service.proto
Wondertan: while serialized can be used in this meaning, the flattened is better suited word here

fibre/client_upload.go
Wondertan: I agree and also wanted to fix this somehow. I am not a fan of the `WithX` approach tho, but don't have a better option atm.

rootulp: Agreed `WithX` isn't ideal because that type may evolve to have more fields.

proto/celestia/fibre/v1/service.proto
```diff
@@ -5,29 +5,44 @@ option go_package = "github.com/celestiaorg/celestia-app/x/fibre/types";
...
   bytes data = 2;
   repeated bytes proof = 3;
-  optional bytes rlc_root = 4; // const len == 32 when present
 }
 
+// Rows represents a collection of rows in a Fibre blob.
+// With two ways of proving data correctness:
+// 1. Using RLC coefficients when there is not enough rows to reconstruct the data blob. (Upload case)
+// 2. Using a Merkle root of RLC tree, when there is sufficient rows to reconstruct the data blob. (Download case)
+message Rows {
```
cmwaters, line 19:
```suggestion
message RowSet {
```
Just a suggestion if you want to avoid Rows.Rows

cmwaters (dismissed):
Makes sense to me

fibre/client_upload.go
cmwaters: `RowSet`?

walldiss (dismissed):

proto/celestia/fibre/v1/service.proto
Wondertan: This makes sense, but it's a very invasive rename. All the request response types, the merged, and my local client and server code all depend on this type, and renaming it will cause me a bunch of pain to update them everywhere atm. Lets rename it once we have all core logic merged with a single PR

walldiss (approved):

PR #30, opened 2025-10-22
feat(fibre): `Server` with `UploadRows` backed by `Store`

Wondertan:
Adds Server skeleton with UploadRows and the Store structure with core storage functionality. The Store uses `go-datastore` interface to abstract over three types of store: Badger, an in-memory one, and Pebble(excluded from the PR). This interface might be removed in later stages once we settle down on the DB backend. 

There are in-progress benchmarks that compare Badger and Pebble performance, allow fine-tuning of their configurations, and demonstrate general server upload performance. All these benchmarks are yet to be concluded in the likely follow-up PR, as they are taking longer than expected. Meanwhile, the core logic is ready for review. 

Includes an integration testing suite with tests for uploads with real clients and servers. Was tested even with 100vals and 100 clients sending 50 blobs each on a dedicated server because my 32GiB on my laptop can't handle this. The committed test is 10, 10, 5 instead.

The tests are non-flaky, and the local linter is happy locally; tracing and logging coverage are at the production level. 

Depends on #29 and https://github.com/celestiaorg/rsema1d/pull/7

### Notable spec Divergencies:
The data model was changed and a bit simplified:
  * Only one "ttl bucket" for both PPs and rows
  * The rows are stored at `<commitment>/<promise_hash>` rather then `<commitment>/<valset_height>`. The older approach requires a `Has` check before `Put` to cover writes for the same commitment and valset_height. This is, however:
     * Forces all the clients' implementations to handle this very rare case gracefully. And handling this case also changes the expectation for the successful upload. If such a collision happens and the client logic reports success, the user would still need to be aware of this case, e.g., they expected to pay for the upload and sign the payment promise, but they didn't.
     * Thrashes the throughput by doing a read before every write.
     * On the contrary, using promise hash does not have the described problems, although in the case of collision with the same height and commitment, the rows will be duplicated. However, the case where two different actors simultaneously upload the same blob is infrequent and thus has negligible overhead.
     
The used grpc errors set is narrower: 
* `ALREADY_EXISTS` is not used as per above  
* `PERMISSION_DENIED` is not used and `INVALID_ARGUMENT` is returned instead for the case.
* `RESOURCE_EXHAUSTED` is not used as it won't be needed once we migrate to [streams](https://github.com/grpc/grpc-go/tree/master/examples/features/flow_control), and will need to, as loading all the data in memory before writing the response is inefficient. 
* `FAILED_PRECONDITION` doesn't exist yet, as there are stateful checks yet, but I also doubt we need it to be this error code.

The flow of the upload logic also diverges a bit with minor changes.


walldiss:
For notable spec divergence we should first discuss it in PRs to the spec repo that describe the change and rational behind it. This will keep spec up to date and would not lead us to unnecessary code changes when not needed. 
 
> Only one "ttl bucket" for both PPs and rows

In implementation you combined storages, however spec defines payments promise storage only for unprocessed PP, not for all. And PP removal logic is complexly different. 

> The rows are stored at <commitment>/<promise_hash> rather then <commitment>/<valset_height>. The older approach requires a Has check before Put to cover writes for the same commitment and valset_height.

How new index remove the need for Has check that supposed to solve deduplication problem. You mention that you want to allow duplication now and it is unclear what is the point of index change

Lets move this discussion to spec PRs pls

walldiss:
> RESOURCE_EXHAUSTED is not used as it won't be needed once we migrate to [streams](https://github.com/grpc/grpc-go/tree/master/examples/features/flow_control), and will need to, as loading all the data in memory before writing the response is inefficient.

This needs to be PRed to spec as well. 
- I'm not aware of plans to migrate to streams, we agreed against it some time ago. 
- Status is supposed to be returned by rate-limiter which is necessary regardless of stream / unary semantics choice

walldiss:
Discussed spec divergence with Hlib in sync and went over rational behind it.

> Only one "ttl bucket" for both PPs and rows

Makes sense, we keep the change. Reason is because it points to the commitment for both indexes and we will skip decoupling PP storage subcomponent for now.

> The rows are stored at /<promise_hash> rather then /<valset_height>. The older approach requires a Has check before Put to cover writes for the same commitment and valset_height.

We kep the change:
- Makes sense for avoiding the case when same commitment+valset_height might be submitted with different timestamp in PP. 
- Deduplication should not be necessary, it will happen on PP validation step. So Has check is not necessary

> RESOURCE_EXHAUSTED is not used as it won't be needed once we migrate to [streams](https://github.com/grpc/grpc-go/tree/master/examples/features/flow_control), and will need to, as loading all the data in memory before writing the response is inefficient.

Streams are not planned. We would need to implement rate-limiting later to protect core-app consensus resources. But thats should be implemented outside of scope of this PR

Wondertan:
@walldiss, will you update the spec accordingly?

fibre/server.go
```diff
@@ -0,0 +1,95 @@
...
+// ServerConfig contains configuration options for the Fibre [Server].
+type ServerConfig struct {
+	// ChainID is the chain identifier for domain separation in [PaymentPromise] validation.
+	ChainID string
+
+	CodingConfig
+
+	// MaxClockDrift is the maximum allowed time difference between the server's clock and the payment promise timestamp.
+	// Promises with timestamps older than (now - MaxClockDrift) will be rejected.
+	MaxClockDrift time.Duration
```
walldiss: It is not clock drift. Clock drift is used to allow difference in clock time on different machines.

fibre/server_upload.go
```diff
@@ -0,0 +1,274 @@
...
+
+	// validate chain ID matches server config
+	if promise.ChainID != s.cfg.ChainID {
+		return nil, nil, fmt.Errorf("payment promisechain ID mismatch: expected %s, got %s", s.cfg.ChainID, promise.ChainID)
+	}
+
+	// validate timestamp is within allowed clock drift
+	now := time.Now()
+	oldestAllowed := now.Add(-s.cfg.MaxClockDrift)
+	if promise.CreationTimestamp.Before(oldestAllowed) {
```
walldiss: We should allow any payment promise that is not timed out yet. So the oldestAllowed should be
``` now.Add(-s.cfg.PromisTimeout-s.cfg.MaxClockDrift) ```

File: fibre/server_upload.go [hunk 3]
```diff
@@ -0,0 +1,274 @@
...
+		return nil, nil, fmt.Errorf("invalid payment promise proto: %w", err)
+	}
+
+	// validate chain ID matches server config
+	if promise.ChainID != s.cfg.ChainID {
+		return nil, nil, fmt.Errorf("payment promisechain ID mismatch: expected %s, got %s", s.cfg.ChainID, promise.ChainID)
+	}
+
+	// validate timestamp is within allowed clock drift
+	now := time.Now()
```
walldiss: Need to compare all times in UTC timezone

fibre/server_upload.go
```diff
@@ -0,0 +1,274 @@
...
+	}, nil
+}
+
+// verifyPromise verifies given proto of [PaymentPromise] and returns unmarshaled form with its hash.
+// It does both stateless and stateful verification.
+func (s *Server) verifyPromise(promisePb *types.PaymentPromise) (*PaymentPromise, []byte, error) {
+	promise := &PaymentPromise{}
+	if err := promise.FromProto(promisePb); err != nil {
+		return nil, nil, fmt.Errorf("invalid payment promise proto: %w", err)
+	}
```
walldiss, line 103: In this function or in verify Assignment need to check that height of promis valset is within bounds. Keep in mind that client potentially can be ahead of server in terms of headerHead, so it need some safety buffer around it

fibre/server_upload.go
```diff
@@ -0,0 +1,274 @@
...
+	// get our validator
+	// TODO(@Wondertan): Should we cache this? It might be worth it for remote KMS and hocrux setups.
+	pubKey, err := s.privVal.GetPubKey()
+	if err != nil {
+		return fmt.Errorf("getting validator public key: %w", err)
+	}
+	ourValidator, found := valSet.GetByAddress(pubKey.Address())
+	if !found {
+		return fmt.Errorf("validator %s not in set at height %d", pubKey.Address().String(), promise.Height)
+	}
```
walldiss: Do we really want to do it on every request or can do it once during init?

fibre/server_upload.go
```diff
@@ -0,0 +1,274 @@
...
+func parseRowSize(rows []*types.Row) (int, error) {
+	if len(rows) == 0 {
+		return 0, errors.New("no rows provided")
+	}
+	rowSize := len(rows[0].Data)
+	if rowSize == 0 {
+		return 0, errors.New("row size cannot be zero")
+	}
+
+	return rowSize, nil
```
walldiss, line 284: might be worth to add upper bound protections for len() calls

fibre/validator/set.go
```diff
@@ -13,6 +15,18 @@ type Set struct {
 	Height uint64
 }
 
+// GetByAddress finds a validator in the set by address.
+// Returns the validator pointer from ValidatorSet.Validators (required for shard map lookups)
+// and true if found, or nil and false if not found.
+func (s Set) GetByAddress(address crypto.Address) (*core.Validator, bool) {
+	for _, val := range s.Validators {
```
walldiss, line 22: Could s.Validators be a map?

walldiss:
Amazing work! I've requested small tweaks here and there. Really good test coverage!

fibre/server.go
Wondertan: > Clock drift is used to allow difference in clock time on different machines.

Isn't this what we do here? The difference in the time.Now on the client and server machines

Wondertan: We should also add the same parameter for height, i.e., we should not allow clients to set any arbitrary heights, as we do currently.

fibre/server_upload.go
Wondertan: Can you elaborate on why? If this is true, then I see where you were coming from with the comment above

fibre/server_upload.go
Wondertan: > Keep in mind that client potentially can be ahead of server in terms of headerHead, so it need some safety buffer around it

While I doubt this can happen, as servers are physically co-located with validators who have the most recent state, the first, this problem can be solved in the `validator.SetGetter` implementation, which would block if the requested height is crrntHead+1

Wondertan: > In this function or in verify Assignment need to check that height of promis valset is within bounds.

AFAIR, there are no height bounds defined in the spec, but we should indeed add it

fibre/server_upload.go
Wondertan: The GetByAddress must be done every time, but getting the pubkey is privval implementation-specific. Imagining how I would implement such an interface over remote KMS, I would def make client-side caching, but it's not me, thus there is todo to check for this.

fibre/validator/set.go
Wondertan: It could be, but consideing that we have only 100 vals, we don't gain much from using it

Wondertan:
> Really good test coverage

There still a few test cases missing for the store itself, but this is on my list.

Also, I still wanna simplify and try using badger's internal TTL, if time permits.

fibre/server_upload.go
Wondertan: I cached just in case anyway

fibre/server_upload.go
Wondertan: done

Wondertan:
Badger TTL thing has some complications that are not worth solving atm, so skipping this.

Everything else I planned for this PR is done and cleaned up. Some of the comments still need responses from @walldiss to be resolved.

Wondertan:
Also, https://github.com/celestiaorg/rsema1d/pull/7 has to be merged still

fibre/server_upload.go
walldiss: We should allow bigger buffer for ts selection, to allow resubmit of PP that error on first attempt for some reason. Because it will force pay and we don't want user to repay on retries. Allowing payment promises that are not timed out is totally valid

Wondertan: It will force pay only if full PP and data were uploaded, i.e. if retry was needed the PP wasn't uploaded and wont be payed for.

fibre/server_upload.go
walldiss: > AFAIR, there are no height bounds defined in the spec, but we should indeed add it

Valiadtion for height bounds might have been accidentally removed, but we should have it. I'll update to add heights bounds to spec if it is missing. We should allow only heights that are within PP timeout window. e.g.
```
heights_bounds = pp_timeout / block_time
```

---
>While I doubt this can happen, as servers are physically co-located with validators who have the most recent state, the first, this problem can be solved in the validator.SetGetter implementation, which would block if the requested height is crrntHead+1

We need to make sure validator.SetGetter implementation actually blocks and not returns "not_found" in such case. And we should allow only small drift (like +1 in example), to prevent locking server on this call forever

fibre/server_upload.go
walldiss: PP can succed on one Val and fail on others. We should allow same PP to be retried within timeout window, not clock drift window

Wondertan: In sync agreed to remove ClockDrift entirely and only keep the PromiseTimeout

fibre/server_upload.go
Wondertan: I really don't like depending on block_time here too, but we have it now

Wondertan:
Ok, https://github.com/celestiaorg/rsema1d/pull/7 has is merged and go.mod updated, need approves

walldiss (approved):

PR #33, opened 2025-10-24
chore: rebase upstream main

evan-forbes:
## Overview

rebases upstream main, this is both just nice to keep up to date but also to pull in the new abci method so that we can use the new promise state that is introduced in https://github.com/celestiaorg/cosmos-sdk/pull/695

blocking celestiaorg/celestia-app#6824


evan-forbes:
we likely need to wait for celestiaorg/celestia-app-fibre#26 to make sure I didn't accidentally miss something

PR #34, opened 2025-10-24
feat(x/fibre): implement msg server

rootulp:
Prerequisite: https://github.com/celestiaorg/fibre-da-spec/pull/27

Part of https://github.com/celestiaorg/celestia-app-fibre/issues/1


x/fibre/keeper/msg_server.go
```diff
@@ -0,0 +1,375 @@
...
+}
+
+// PaymentPromiseTimeout processes a payment promise after the timeout period
+func (ms msgServer) PaymentPromiseTimeout(goCtx context.Context, msg *types.MsgPaymentPromiseTimeout) (*types.MsgPaymentPromiseTimeoutResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// Validate the message
+	if err := msg.ValidateBasic(); err != nil {
+		return nil, err
+	}
```
evan-forbes: not blocking but there's an easy fix here

pretty sure these are called in the antehandlers https://github.com/celestiaorg/celestia-app/blob/fc8cd9d1909f09ad3208eebfde7874ef5f58dd3e/app/ante/ante.go#L40

so we don't want to re-verify them here

this was done in the other methods, but I only leave this single comment.

x/fibre/keeper/msg_server.go
```diff
@@ -0,0 +1,375 @@
...
+
+	// Validate payment promise internally
+	if err := ms.ValidatePaymentPromiseInternal(ctx, &msg.PaymentPromise); err != nil {
+		return nil, errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "payment promise validation failed: %s", err)
+	}
+
+	// Check if payment promise has already been processed
+	if ms.IsPaymentPromiseProcessed(ctx, &msg.PaymentPromise) {
+		return nil, errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "payment promise has already been processed")
+	}
```
evan-forbes: definitely not blocking

pretty unimportant, but we're converting to the internal version here in validatepaymentpromiseinternal and ispaymentpromiseProcessed

most of the work besides the actual signature verification for ValidatePaymentPromiseInternal is being done in the validate basic call

x/fibre/keeper/msg_server.go
```diff
@@ -0,0 +1,375 @@
...
+
+	// Validate validator signatures using existing fibre/validator.SignatureSet
+	if err := ms.validateValidatorSignatures(ctx, &msg.PaymentPromise, msg.ValidatorSignatures); err != nil {
+		return nil, errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "validator signature validation failed: %s", err)
+	}
```
evan-forbes: definitely not blocking

generally it feels like these are taking to high level of args but could be wrong.

for example all we need here are the sign bytes, height, and validator signatures. personally, that helps me read the function as limiting what goes it clarifies what this function is doing

it also allows us to get past the minor annoyance above where we're repeated a lot of work (easy work not worth optimizing atm, but just makes the logic a bit murkier imo

x/fibre/keeper/msg_server.go
```diff
@@ -0,0 +1,375 @@
...
+
+	// Check if sufficient balance (should always be true since promise was validated, but safety check)
+	if escrowAccount.Balance.IsLT(paymentAmount) {
+		return nil, errorsmod.Wrapf(sdkerrors.ErrInsufficientFunds, "insufficient balance: have %s, need %s", escrowAccount.Balance, paymentAmount)
+	}
+
+	// Deduct payment from escrow account (both balance and available_balance)
+	escrowAccount.Balance = escrowAccount.Balance.Sub(paymentAmount)
+	escrowAccount.AvailableBalance = escrowAccount.AvailableBalance.Sub(paymentAmount)
+	ms.SetEscrowAccount(ctx, escrowAccount)
```
evan-forbes, line 250: not blocking

do you think we could isolate the stateful verification of the payment promise? then we could use that method here, in the PFF, and when we are verifying the promise itself before signing

x/fibre/keeper/msg_server.go
```diff
@@ -0,0 +1,375 @@
...
+
+	// Check if sufficient available balance
+	if escrowAccount.AvailableBalance.IsLT(paymentAmount) {
+		return nil, errorsmod.Wrapf(sdkerrors.ErrInsufficientFunds, "insufficient available balance: have %s, need %s", escrowAccount.AvailableBalance, paymentAmount)
+	}
+
+	// Deduct payment from escrow account
+	escrowAccount.Balance = escrowAccount.Balance.Sub(paymentAmount)
+	escrowAccount.AvailableBalance = escrowAccount.AvailableBalance.Sub(paymentAmount)
+	ms.SetEscrowAccount(ctx, escrowAccount)
```
evan-forbes, line 175: what happens if something is wrong here and the balance goes negative?

evan-forbes (commented):
where are the msg server tests? are we waiting to wire up the application or perhaps I missed them?

x/fibre/keeper/msg_server.go
rootulp: Sorry I don't understand what's the issue or proposed fix?

x/fibre/keeper/msg_server.go
rootulp: Is the concrete proposal to do this?

```diff
-	if err := ms.validateValidatorSignatures(ctx, &msg.PaymentPromise, msg.ValidatorSignatures); err != nil {
+	pp := fibre.PaymentPromise{}
+	if err := pp.FromProto(&msg.PaymentPromise); err != nil {
+		return nil, errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "failed to convert payment promise: %s", err)
+	}
+	signBytes, err := pp.SignBytes()
+	if err != nil {
+		return nil, errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "failed to get sign bytes: %s", err)
+	}
+	if err := ms.validateValidatorSignatures(ctx, signBytes, &msg.PaymentPromise.Height, msg.ValidatorSignatures); err != nil {
```

rootulp: 1792c14fb10f615c6e7d126dd43a9b3bc2858b66

x/fibre/keeper/msg_server.go
rootulp: d79411df16fa975031beb4d228eca8d1ea7248ad

x/fibre/keeper/msg_server.go
rootulp: I added a check to verify that balance can't go negative. 0c09b464f50e0d2b06cde6da9d76b18c94ffd53a

x/fibre/keeper/msg_server.go
rootulp: TBH I liked the previous implementation more because IMO fewer functions arguments means less cognitive load + improved readability.

But will defer to reviewer for things like this to get it merged.

x/fibre/keeper/msg_server.go
rootulp: b6749a07a11ca4c019b70de41e95ca571ce80edb

rootulp:
> where are the msg server tests? are we waiting to wire up the application or perhaps I missed them?

Just added them in [82d0d55](https://github.com/celestiaorg/celestia-app-fibre/pull/34/commits/82d0d55bfc2760cf5d03013e8cf9e2e2243338bc)

x/fibre/keeper/msg_server.go
evan-forbes: we're just doing a bit or redundant unmarshalling

we could unmarshal once and then verify the signature

ValidatePaymentPromiseInternal and IsPaymentPromiseProcessed are both unmarshalling the promise

evan-forbes (dismissed):
pre-approving

evan-forbes (approved):

PR #40, opened 2025-10-29
WIP: promise bank

evan-forbes:
WIP not ready for review

just posting existing code


evan-forbes:
unsure if we'll keep this if block times go down and this has to be so outdated might as well start from scratch

generally the approach doesn't seem horrible, but it adds a meaningful amount of complexity

PR #39, opened 2025-10-28
feat(x/fibre): add `available_timestamp` to withdrawal

rootulp:
Implements the spec change in https://github.com/celestiaorg/fibre-da-spec/pull/29

Closes https://github.com/celestiaorg/celestia-app-fibre/issues/31 but I'm extracting the `BeginBlock` iteration to a separate PR.

PR #41, opened 2025-10-29
feat(x/fibre): process withdrawals in `BeginBlock`

rootulp:
~Blocked on https://github.com/celestiaorg/celestia-app-fibre/pull/39~

Processes withdrawals in `BeginBlock`. There will be a follow-up PR to add more logic to `BeginBlock` specifically to prune payment promises.


evan-forbes (dismissed):
lgtm

x/fibre/keeper/abci_test.go
```diff
@@ -0,0 +1,347 @@
...
+	suite.Equal(requestTime, withdrawals[0].RequestedTimestamp)
+
+	// CHANGE THE PARAMETER: Update withdrawal delay to 48 hours
+	params := suite.keeper.GetParams(suite.ctx)
+	originalDelay := params.WithdrawalDelay
+	params.WithdrawalDelay = 48 * time.Hour
+	suite.keeper.SetParams(suite.ctx, params)
+
+	// Advance time to original availableAt (T + 24h)
+	// The withdrawal should still be processed because we stored the actual RequestedTimestamp
```
cmwaters, line 299: I thought it's the `AvailableTimestamp` not the `RequestedTimestamp` which is the reason why changes to the params won't affect existing withdrawals

cmwaters (approved):
Looks good.

I couldn't find the test. If we don't have one yet can we test that if you have multiple withdrawals but not enough balance the second withdrawal isn't executed.

x/fibre/keeper/abci_test.go
rootulp: Agreed looks like it should be AvailableTimestamp. Will fix and add a test in a FLUP.

PR #49, opened 2025-10-30
feat(fibre): `client.Put` sends PFF

Wondertan:
As PFF is now available, adding the missing submission piece to the `Put` method.


Wondertan:
Open to ideas on how to test this

fibre/client_put.go
```diff
@@ -62,13 +59,51 @@ func (c *Client) Put(ctx context.Context, ns share.Namespace, data []byte) (resu
...
+		return result, fmt.Errorf("confirming PayForFibre transaction: %w", err)
+	}
 
-	// TODO(@Wondertan): Submit and await inclusion of MsgPayForFibre transaction through txClient.
+	if txResp.Code != 0 {
+		err := fmt.Errorf("PayForFibre transaction failed with code %d", txResp.Code)
+		span.RecordError(err)
+		span.SetStatus(codes.Error, "PayForFibre transaction failed")
+		return result, err
+	}
```
cmwaters: You can consolidate this into

`c.txClient.SubmitTx`

Also you don't need to check the txResp.Code, the tx client should do that for you and return the error

fibre/client_put.go
cmwaters, line 33: Eventually you might want to expose tx options like we have in the light node etc. but this is not necessary for now I don't think

cmwaters (commented):
Just a quick passthrough although there's not much needed here. I can help with the e2e testing if you want

fibre/client_put.go
Wondertan: I intentionally separated them to have different trace events for each. Noted on the response code.

fibre/client_put.go
Wondertan: Yeah, we will have it. We first need to see how it will play out with node's API

Wondertan:
> I can help with the e2e testing if you want

Thank you. I guess we will need to connect server to app first to make it e2e which will be out of scope of this PR

walldiss (commented):
Would it be possible to use TxClient for that?

walldiss:
Testing might be easier if the PFF submission logic is isolated into dedicated call, which will be possible to test in isolation from the rest of the Put logic

Wondertan:
> Would it be possible to use TxClient for that?

It is used already

Wondertan:
> Testing might be easier if the PFF submission logic is isolated into dedicated call, which will be possible to test in isolation from the rest of the Put logic

The thing is, there is no PFF submission logic, it just creates the struct and passes it to txClient

walldiss (dismissed):

cmwaters:
> > Testing might be easier if the PFF submission logic is isolated into dedicated call, which will be possible to test in isolation from the rest of the Put logic
> 
> The thing is, there is no PFF submission logic, it just creates the struct and passes it to txClient

Yeah I agree

cmwaters (dismissed):

walldiss (approved):

PR #50, opened 2025-10-30
feat(fibre/validator): GrpcSetGetter

Wondertan:
Allows fetching validator sets over GRPC. 

Handling the case where the requested height is head+1 is out of scope of this PR and should be handled on the core side.

### DIFF
--- a/fibre/validator/grpc_set_getter.go
+++ b/fibre/validator/grpc_set_getter.go
@@ -1,0 +1,66 @@
+package validator
+
+import (
+	"context"
+	"fmt"
+
+	coregrpc "github.com/cometbft/cometbft/rpc/grpc"
+	core "github.com/cometbft/cometbft/types"
+)
+
+// GrpcGetter implements the [SetGetter] interface by fetching [Set]
+// using the BlockAPI gRPC client.
+type GrpcGetter struct {
+	client coregrpc.BlockAPIClient
+}
+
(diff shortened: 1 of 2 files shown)


rootulp:
I observed that CI fails on your PRs but it passes on mine. 

Example: https://github.com/celestiaorg/celestia-app-fibre/pull/48

My PRs use branches from my fork but your PRs use branches on the canonical repo. Not sure if that has something to do with it.

Wondertan:
No idea why is that the case honestly

walldiss (approved):

fibre/validator/grpc_set_getter.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+
+	coregrpc "github.com/cometbft/cometbft/rpc/grpc"
+	core "github.com/cometbft/cometbft/types"
+)
+
+// GrpcGetter implements the [SetGetter] interface by fetching [Set]
+// using the BlockAPI gRPC client.
+type GrpcGetter struct {
+	client coregrpc.BlockAPIClient
```
cmwaters, line 14: Interesting in my tests, I used cmtservice.ServiceClient. Seems like there's multiple ways to get the validator set. This one looks actually better because there is no pagination you need to worry about and gets the core.ValidatorSet endpoint. I will update my tests to use this

cmwaters (approved):

PR #52, opened 2025-10-30
feat(fibre): grpc host registry

cmwaters:
Connects with the valaddr module on-chain to provide information on the fibre host names.

I'm open to ideas but my approach in my head is to use `GetHost()` and if the client isn't unable to send the row to that particular validator because their is no response from the host, use `PullHost()` to see if it's been updated since it was last cached and if the address is different, try again.


fibre/grpc/host_registry.go
```diff
@@ -0,0 +1,79 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+var _ validator.HostRegistry = &HostRegistry{}
+
+// HostRegistry is a registry of validator hosts. It caches the hosts for validators in the active set.
+// It uses the [types.QueryClient] to query the fibre provider information for validators in the active set.
+type HostRegistry struct {
+	queryClient types.QueryClient
+	cachedHosts map[string]validator.Host //
```
rootulp:
```suggestion
	cachedHosts map[string]validator.Host
```

rootulp (dismissed):
LGTM, makes sense

fibre/grpc/host_registry.go
```diff
@@ -0,0 +1,79 @@
+package grpc
```
Wondertan, line 1: Let's move this to `fibre/grpc`, maybe you could also move the `fibre/validator.GrpcGetter` there as `fibre/grpc.SetGetter`.

fibre/grpc/host_registry.go
```diff
@@ -0,0 +1,79 @@
...
+	if len(g.cachedHosts) == 0 {
+		if err := g.PullAll(ctx); err != nil {
+			return "", err
+		}
+		if host, ok := g.cachedHosts[valConAddr]; ok {
+			return host, nil
+		} else {
+			return "", fmt.Errorf("host not found for validator %s", valConAddr)
+		}
+	}
```
Wondertan, line 65: Instead of doing a lazy pull that causes unexpected latency spikes, we could also just make a Start method that initiates the registry by pulling the hosts. Start would be expected to block.

fibre/grpc/host_registry.go
```diff
@@ -0,0 +1,79 @@
...
+		ValidatorConsensusAddress: consAddr.String(),
+	})
+	if err != nil {
+		return "", err
+	}
+	if !resp.Found {
+		return "", fmt.Errorf("host not found for validator %s", consAddr.String())
+	}
+	g.cachedHosts[val.Address.String()] = validator.Host(resp.Info.Host)
+	return validator.Host(resp.Info.Host), nil
```
Wondertan: If we don't do any validation on the state side(which is still ridiculous to me btw), we should verify it here at least. Domain names also have formatting. Also, the Host type is expected to have "tcp://" in it to be dialable by grpc.

fibre/grpc/host_registry.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+	}
+
+	// look up the specific validator's host if it's missing from the cache. It might have
+	// been added to the active set since the last refresh.
+	return g.PullHost(ctx, val)
+}
+
+// PullAll pulls all active fibre providers from the query client and caches them, overwriting any existing cached hosts.
+func (g *HostRegistry) PullAll(ctx context.Context) error {
```
Wondertan, line 73: optional uber nit: not a native speaker, but usually the word fetch is used instead of pull.

Wondertan (commented):
> I'm open to ideas but my approach in my head is to use GetHost() and if the client isn't unable to send the row to that particular validator because their is no response from the host, use PullHost() to see if it's been updated since it was last cached and if the address is different, try again.

I am not sure why we would use subscription instead, but if we go this path, we should also update `fibre/grpc.DefaultNewClientFn` to have this logic to fetch hostname on error. One issue with this approach is that additional fetching may cause unexpected latency spikes for the user. It may be fine for now, but one could also argue that in the future, when we forget about this, it's gonna take some time to figure out where the latency is coming from.

fibre/grpc/host_registry.go
cmwaters: Is there any other operation we'd like to do at start. The reason I stalled is to make sure that the client pulls the latest version of the registry

fibre/grpc/host_registry.go
cmwaters: git pull or git fetch?

Wondertan: hehe

fibre/grpc/host_registry.go
Wondertan: I mean, particularly the PullAll case when there are no cachedHosts, which can only happen on the first start, which is better off in the Start method, where blocking is expected.

fibre/grpc/host_registry.go
```diff
@@ -0,0 +1,96 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+var _ validator.HostRegistry = &HostRegistry{}
+
+// HostRegistry is a registry of validator hosts. It caches the hosts for validators in the active set.
+// It uses the [types.QueryClient] to query the fibre provider information for validators in the active set.
+type HostRegistry struct {
+	queryClient types.QueryClient
+	cachedHosts map[string]validator.Host
```
Wondertan, line 22: We should synchronize access to it, including single flight protection for fetches. There can be multiple Uploads that access the registry

fibre/grpc/set_getter.go
```diff
@@ -23,44 +24,44 @@ func NewGrpcGetter(client coregrpc.BlockAPIClient) *GrpcGetter {
 
 // Head returns the latest [Set] by calling getByHeight with 0 height.
 // This avoids an additional roundtrip to get the status first.
-func (g *GrpcGetter) Head(ctx context.Context) (Set, error) {
+func (g *GrpcGetter) Head(ctx context.Context) (validator.Set, error) {
```
Wondertan: Lets rename it to SetGetter. grpc is in the pkg already.

fibre/grpc/host_registry_test.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+func TestIntegrationTestSuite(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping grpc host registry test in short mode.")
+	}
+	suite.Run(t, &IntegrationTestSuite{})
+}
```
Wondertan: This is a bit incosistent with how tests are structured within fibre pkg, but actually, that's not a big deal atm

fibre/grpc/host_registry.go
```diff
@@ -0,0 +1,96 @@
...
+	host, err := g.getHost(ctx, val)
+	if err != nil {
+		return "", err
+	}
+
+	// check if the host is a valid URL
+	_, err = url.Parse(host.String())
+	if err != nil {
+		return "", fmt.Errorf("got invalid host %s: %w", host.String(), err)
+	}
```
Wondertan, line 42: Shouldn't we verify it immediately after Pull? Before setting an invalid value in the cache?

Wondertan: We should also unit test this to ensure we have the expected address format + some error cases.

Wondertan (commented):
>  if the client isn't unable to send the row to that particular validator because their is no response from the host, use PullHost() to see if it's been updated since it was last cached and if the address is different, try again.

As we agreed in sync, we want to have this logic anyway, so we should add it within this PR. On a quick look, it should land nicely in `grpc.DefaultNewClientFn`

fibre/grpc/host_registry.go
cmwaters: Got it, didn't realise it needed to be concurrent safe

fibre/grpc/host_registry.go
cmwaters: If we don't leave the invalid value in the cache then we will pull every time GetHost is called. 

I don't want all these extra round trips just because someone uploaded an invalid or unreachable host. 

Let me think up a solution to this

fibre/grpc/host_registry_test.go
cmwaters: I don't think we have any of these integrations tests yet which I think for this component is important. I should also add some unit tests

fibre/grpc/host_registry.go
cmwaters: I didn't realise there was a `Start` method - it isn't part of the interface

fibre/grpc/host_registry.go
cmwaters: I think we should leave it as is to avoid doing potentially multiple network request everytime we have a poorly constructed host. 

Long term, we may be better off with a subscription service. I will try write this down as an issue

fibre/grpc/host_registry.go
Wondertan: There is no, and I am proposing to add it.

fibre/grpc/host_registry_test.go
Wondertan: We have integration tests for client_server

fibre/grpc/host_registry.go
```diff
@@ -0,0 +1,126 @@
...
+	queryClient types.QueryClient
+	mu          sync.RWMutex
+	cachedHosts map[string]validator.Host
+}
+
+func NewHostRegistry(queryClient types.QueryClient) *HostRegistry {
+	return &HostRegistry{
+		queryClient: queryClient,
+		cachedHosts: make(map[string]validator.Host),
+	}
```
rootulp, line 29: [question] why doesn't this create a mutex and populate `mu`?

rootulp (approved):
modernize-check seems to be failing

Wondertan:
@cmwaters, this should not be merged without @walldiss or mine's approval... At least i had unresolved comment and some pending ones.

fibre/grpc/host_registry.go
cmwaters: It doesn't need to I don't think

cmwaters:
> @cmwaters, this should not be merged without @walldiss or mine's approval... At least i had unresolved comment and some pending ones.

Sorry, I think I had auto merge on because I was just waiting on your review. What unresolved comments did you have. I can address them as FLUPs

PR #53, opened 2025-10-30
feat(fibre): downloading

Wondertan:
## Summary

  Implements `Client.Download` for retrieving and reconstructing blobs from the Fibre network.

  ### Algorithm

  Downloads use a stake-weighted validator selection strategy optimized for the happy path:

  1. **Minimal initial request set**: Select only the minimum required validators (typically ~1/3 by stake) using `Select`, shuffled by stake weight for load balancing
  2. **Happy path**: If all initial requests succeed, reconstruction completes in a single round-trip
  3. **Fault tolerance**: If any request fails, automatically issue recovery requests to additional validators until either reconstruction succeeds or all validators are exhausted

  This approach minimizes network overhead in the common case while maintaining resilience against validator failures.

  ### Test Infrastructure

  - Consolidated upload and download integration tests to reduce test runtime
  - Unified mock implementations between upload and download tests
  - Added coverage for all known edge cases: success, fault tolerance with various failure counts, context cancellation, and closed client handling
  
 
### TODO/FLUPS: 
* download benchmarks. 
* Move reconstruction logic to rsema1d https://github.com/celestiaorg/rsema1d/issues/14

Depends on #156


fibre/client_download.go
```diff
@@ -0,0 +1,276 @@
...
+		}
+		proofs = append(proofs, &rsema1d.RowInclusionProof{
+			RowProof: rsema1d.RowProof{
+				Index:    int(row.Index),
+				Row:      row.Data,
+				RowProof: row.Proof,
+			},
+			RLCRoot: rlcRoot,
+		})
+	}
```
walldiss, line 300: Need to check row.Index < max row index. Or maybe I just don't see where this check happens. Just to make sure when populating rows we don't get panic

fibre/blob.go
```diff
@@ -189,6 +213,96 @@ func (d *Blob) Row(index int) (*rsema1d.RowInclusionProof, error) {
...
+	case collected >= d.cfg.OriginalRows:
+	}
+
+	// use reedsolomon decoder directly as opposed to rsema1d.Reconstruct
+	// the decoder is used to reconstruct missing shards in-place which is more efficient than copying data and
+	// passing indicies as a slice of integers.
+	enc, err := reedsolomon.New(d.cfg.OriginalRows, d.cfg.ParityRows, reedsolomon.WithLeopardGF16(true))
+	if err != nil {
+		return fmt.Errorf("creating reedsolomon decoder: %w", err)
+	}
```
walldiss, line 254: from rsmt2d I remember that encoder needs to be cached in memory to allow much more efficient reconstruction. It was giving massive performance important in rsmt2d. We should avoid recreating it every time

fibre/blob.go
```diff
@@ -189,6 +213,96 @@ func (d *Blob) Row(index int) (*rsema1d.RowInclusionProof, error) {
...
+	// TODO(@Wondertan): Move and encapsulate inside rsema1d
+
+	// check if we have enough rows
+	collected := int(d.rowsCollected.Load())
+	switch {
+	case collected == 0:
+		return ErrBlobNotFound
+	case collected < d.cfg.OriginalRows:
+		return fmt.Errorf("%w: collected %d rows, need %d", ErrNotEnoughRows, collected, d.cfg.OriginalRows)
+	case collected >= d.cfg.OriginalRows:
```
walldiss: nit:this case condition can be just replaced with default

fibre/blob.go
Wondertan: This is also how rsema does it. I think we can tackle this together with https://github.com/celestiaorg/rsema1d/issues/14

fibre/client_download.go
Wondertan: it is checked inside rsema

fibre/client_download.go
```diff
@@ -0,0 +1,315 @@
...
+		return err
+	}
+	span.AddEvent("client_acquired")
+
+	resp, err := client.DownloadShard(ctx, &types.DownloadShardRequest{Commitment: commitment[:]})
+	if err != nil {
+		if context.Cause(ctx) == errDownloaded {
+			span.SetStatus(codes.Ok, "")
+			return err
+		}
```
walldiss, line 122: could you elaborate on what is this condition and why it is OK, while returning error. Also wonder why error parsed from context and not by `error.Is` check which is more ideomatic

fibre/client_download_test.go
```diff
@@ -0,0 +1,271 @@
...
+	}, blob)
+	defer client.Close()
+
+	downloaded, err := client.Download(t.Context(), blob.Commitment())
+	require.NoError(t, err)
+	require.Equal(t, blob.Data(), downloaded.Data())
+
+	// downloadTarget = 10 * 2/3 = 6
+	// we should have exactly 6 successful downloads (no over-fetching in happy path)
+	require.Equal(t, int64(6), counter.Load(), "should download from exactly downloadTarget validators")
```
walldiss: Given encoding is 1 original to 3 parity it should be enough to download from 1/4+1 of val set in happy path. Perhaps happy path `max(...)` calculation can be more optimistic than it currently is.

fibre/client_download.go
```diff
@@ -0,0 +1,315 @@
...
+
+	var (
+		successes    atomic.Uint32         // tracks successful responses
+		downloadedCh = make(chan struct{}) // closes when downloadTarget amount of responses complete
+	)
+
+	var (
+		// limit to download minimum required amount of shards in the best case, instead of everything
+		downloadTarget = max(valSet.Size()*int(c.cfg.UploadTargetSignaturesCount.Numerator)/
+			int(c.cfg.UploadTargetSignaturesCount.Denominator), 1)
```
walldiss: Re another comment. downloadTarget should be based on erasure encoding factor (1:3) and not signature count. So should be val_set_size*n/(n+k)+1

fibre/client_download.go
Wondertan: See how errDownloaded is used, this will answer all the questions regarding it

fibre/client_download.go
Wondertan: Yes, I was talking to nash, and he mentioned this too. Currently looking into reworking the configuration to accommodate this and other parametrizations

fibre/blob.go
Wondertan: Actually, this is also part of the row verification codepath, and it was never visible profiles

fibre/validator/set.go
```diff
@@ -87,6 +88,74 @@ func (s Set) Assign(commitment rsema1d.Commitment, totalRows, minRows int) Shard
...
+		return nil, 0
+	}
+
+	validators = make([]*core.Validator, len(s.Validators))
+	copy(validators, s.Validators)
+
+	// find split between non-overlapping (top 2/3+1 stake) and potentially overlapping
+	// NOTE: overlapping validators may get duplicate row and the goal here is to prevent them
+	// from being selected within the minRequired bounds. They still will be used as backup download sources.
+	nonOverlapStake := s.TotalVotingPower()*2/3 + 1
```
Wondertan: cc @nashqueue

nashqueue: It should be something like the code below and would you return the last validator who does not have overlapping stake. 

for i = 0 , i++ , in validators
if  i.stake > 1,2 %
    add i.stake
else
   add 1,2%
if total > 100%
    break and return i-1

The 1.2 % comes from the minimum samples being 148 samples for UD. 
4096*3 is the total number of distributed rows required so 33% can reconstruct. 

148/(4096*3 ) ~ 1.2%

nashqueue:
In the Assign function, you need to look at the following relation: 

4069 / X = 33,33% / stake% 

33% of the stake should have 4096 rows.  Therfore how many rows do i need for stake%

X=ceil((4096/33,33% * stake%)

fibre/validator/set.go
Wondertan: Resolved in #161

walldiss (dismissed):

walldiss (approved):

PR #59, opened 2025-11-03
feat: start Fibre `server` on celestia-appd start

### LINKED ISSUE #51: Wire server up to consensus node
## Context

Hlib is implementing a Fibre server in https://github.com/celestiaorg/celestia-app-fibre/pull/30 

## Problem

Currently the server code isn't hooked up to `celestia-appd start` so consensus nodes (and specifically validators) aren't running the Fibre server.

## Proposal

Hook up the Fibre server to `celestia-appd start`. 

Note: any params that the validator needs to specify should be specified via CLI flags initially. At a later point, we can add a config file for them.

I think this is blocked on https://github.com/celestiaorg/celestia-app-fibre/pull/30

rootulp:
Closes https://github.com/celestiaorg/celestia-app-fibre/issues/51

Summary:
- Rename `modify_root_command.go` to `modify_root_command_standalone.go` b/c it only applies to the standalone binary. One change in there to use a `startCommandHandler` that now starts a Fibre server
- `start_command_standalone.go` is mostly AI generated but based on the multiplexer start command.

## Testing

Setup

```shell
make install
./scripts/single-node.sh
```

Help docs

```shell
celestia-appd start --help # shows Fibre help
```

GRPC

```shell
grpcurl -plaintext localhost:9090 list | grep -i fibre
celestia.fibre.v1.Fibre
celestia.fibre.v1.Query

grpcurl -plaintext localhost:9090 list celestia.fibre.v1.Fibre
celestia.fibre.v1.Fibre.DownloadRows
celestia.fibre.v1.Fibre.UploadRows

grpcurl -plaintext -d '{"commitment": "0000000000000000000000000000000000000000000000000000000000000000"}'   localhost:9090 celestia.fibre.v1.Fibre/DownloadRows
ERROR:
  Code: Unimplemented
  Message: method DownloadRows not implemented
```

Repeated with the standalone binary

## FLUPs

- [x] https://github.com/celestiaorg/celestia-app-fibre/issues/61
- [ ] https://github.com/celestiaorg/celestia-app/issues/6816


rootulp:
I hit this:

```
celestia-app-fibre rp/wire-up-consensus-node $ ./scripts/single-node.sh
celestia-app version: cb927640
celestia-app home: /Users/rootulp/.celestia-app
celestia-app genesis file: /Users/rootulp/.celestia-app/config/genesis.json

Do you want to delete existing /Users/rootulp/.celestia-app and start a new local testnet? [y/n]
y
Deleting /Users/rootulp/.celestia-app...
Initializing validator and node config files...
Adding a new key to the keyring...
Adding genesis account...
Creating a genesis tx...
Collecting genesis txs...
Tracing is set up with the ability to pull traced data from the node on the address http://127.0.0.1:26661
Starting celestia-app...
4:17PM INF P2P Node ID ID=1e84ea56ec31172ffdcf6a73672ce9d9c9d19894 file=/Users/rootulp/.celestia-app/config/node_key.json module=p2p
4:17PM INF Adding persistent peers addrs=[] module=p2p
4:17PM INF Adding unconditional peer ids ids=[] module=p2p
4:17PM INF Add our address to book addr=1e84ea56ec31172ffdcf6a73672ce9d9c9d19894@0.0.0.0:26656 book=/Users/rootulp/.celestia-app/config/addrbook.json module=p2p
4:17PM INF service start impl="P2P Switch" module=p2p msg="Starting P2P Switch service"
4:17PM INF service start book=/Users/rootulp/.celestia-app/config/addrbook.json impl=AddrBook module=p2p msg="Starting AddrBook service"
2025/11/03 16:17:20 FATAL: [core] grpc: Server.RegisterService after Server.Serve for "celestia.fibre.v1.Fibre"
```

cmd/celestia-appd/cmd/fibre_server.go
```diff
@@ -0,0 +1,157 @@
...
+
+	if !isValidator {
+		svrCtx.Logger.Info("Node is not a validator, skipping Fibre server startup")
+		return nil, nil
+	}
+
+	svrCtx.Logger.Info("Initializing Fibre server for validator node")
+
+	// Get PrivValidator from CometBFT node
+	privVal := cmtNode.PrivValidator()
```
rootulp: TODO create a FLUP to investigate if this works for horcrux and tmkms.

cmd/celestia-appd/cmd/root.go
```diff
@@ -39,6 +40,15 @@ const (
...
 	DelayedPrecommitTimeoutFlag = "delayed-precommit-timeout"
+
+	// FibreEnableFlag enables or disables the Fibre server
+	FibreEnableFlag = "fibre.enable"
+	// FibreStoreTypeFlag specifies the store type: "memory" or "badger"
+	FibreStoreTypeFlag = "fibre.store-type"
+	// FibreStorePathFlag specifies the path for the badger store
+	FibreStorePathFlag = "fibre.store-path"
+	// FibreBlockTimeFlag specifies the expected block time
+	FibreBlockTimeFlag = "fibre.block-time"
```
rootulp: todo: create a follow-up to remove this flag from the Fibre server. We can't guarantee block time so this will likely be out of sync with the true block time so it can't be used for any critical operations.

rootulp: https://github.com/celestiaorg/celestia-app-fibre/issues/60

cmd/celestia-appd/cmd/fibre_server.go
```diff
@@ -0,0 +1,158 @@
...
+	"google.golang.org/grpc"
+)
+
+// isValidatorNode checks if the node has a PrivValidator configured.
+// Returns true if PrivValidatorKeyFile exists and is readable, false otherwise.
+func isValidatorNode(cfg *cmtcfg.Config) bool {
+	pvKeyFile := cfg.PrivValidatorKeyFile()
+	_, err := os.Stat(pvKeyFile)
+	return err == nil
+}
```
cmwaters: Non validators also can have this File PV setup. Also validators could also be using a KMS which means they might not have a file there (I don't actually know the TMKMS implementation too well)

cmd/celestia-appd/cmd/fibre_server.go
```diff
@@ -0,0 +1,158 @@
...
+	// Get chain ID from config or genesis (should match the node's chain ID)
+	chainID := svrCtx.Viper.GetString("chain-id")
+	if chainID == "" {
+		// Fallback: try to get chain ID from genesis
+		genDoc := cmtNode.GenesisDoc()
+		if genDoc != nil {
+			chainID = genDoc.ChainID
+		} else {
+			// Default fallback
+			chainID = appconsts.MainnetChainID
```
cmwaters: Can we just error here instead of assuming that it is mainnet

cmd/celestia-appd/cmd/fibre_server.go
```diff
@@ -0,0 +1,158 @@
...
+	if storeType == "" {
+		storeType = "badger" // default
+	}
+
+	storeCfg := fibre.DefaultStoreConfig()
+	var store *fibre.Store
+	var err error
+
+	switch storeType {
+	case "memory":
```
cmwaters: do we have constants for these?

cmd/celestia-appd/cmd/fibre_server.go
```diff
@@ -0,0 +1,158 @@
...
+	fibreEnabled := svrCtx.Viper.GetBool(FibreEnableFlag)
+	if !fibreEnabled {
+		svrCtx.Logger.Info("Fibre server is disabled via flag, skipping startup")
+		return nil, nil
+	}
+
+	if !isValidator {
+		svrCtx.Logger.Info("Node is not a validator, skipping Fibre server startup")
+		return nil, nil
+	}
```
cmwaters: Can this be moved outside the function - seems unnatural like this

cmd/celestia-appd/cmd/start_command_standalone.go
```diff
@@ -0,0 +1,351 @@
...
+func startCommandHandler(
+	svrCtx *server.Context,
+	clientCtx client.Context,
+	appCreator servertypes.AppCreator,
+	withCmt bool,
+	opts server.StartCmdOptions,
+) error {
+	if !withCmt {
+		return fmt.Errorf("cannot start app without CometBFT")
+	}
```
cmwaters, line 47: Can this check be made outside the function

cmd/celestia-appd/cmd/fibre_server.go
cmwaters: Also I think you should be able to start a fibre node regardless of whether you are a validator or not

cmd/celestia-appd/cmd/root.go
cmwaters: Why do we depend on block time in the first place - I haven't seen it before

multiplexer/abci/fibre_server.go
```diff
@@ -0,0 +1,153 @@
...
+	pvKeyFile := cfg.PrivValidatorKeyFile()
+	_, err := os.Stat(pvKeyFile)
+	return err == nil
+}
+
+// startFibreServer initializes and registers the Fibre server with the gRPC server for a validator node.
+// Returns the Fibre server instance and an error. The server should be stopped gracefully during shutdown.
+// If isValidator is false, this function does nothing and returns nil, nil.
+// If isValidator is true and initialization fails, returns an error (preventing node startup).
+func (m *Multiplexer) startFibreServer(
```
cmwaters: Do we have two copies of this method so it can be run in multiplexer mode and standalone mode (for a single version). Is there a way we can consolidate those methods

cmwaters (commented):
Good start - left a few comments

cmd/celestia-appd/cmd/fibre_server.go
rootulp: Good catch. Tried to address in 1141e0b00b4377fd3413226800e08e7ce35f6bf2 but will also add a TODO to verify this works for tmkms.

multiplexer/abci/fibre_server.go
rootulp: Yep, and agreed. d9cd7ea6f40beff530d48053ad2ddf34505434e2

cmd/celestia-appd/cmd/root.go
rootulp: Seems like for max height drift which I assume will be error prone: https://github.com/celestiaorg/celestia-app-fibre/pull/30/files#diff-697a846a5a94db6984569ccb695d515d3a9d451ec57ca703476907546de59d57R125-R128

Ref: https://github.com/celestiaorg/celestia-app-fibre/pull/30 cc: @Wondertan

cmd/celestia-appd/cmd/fibre_server.go
rootulp: Good idea. 17e1bc09e70cee435656ae80431e30c3c1abeafd

cmd/celestia-appd/cmd/fibre_server.go
rootulp: Agreed it was jank. Deleted the default chainID thing in ae28e42731150aa89a97c083d60e8355245e2974

cmd/celestia-appd/cmd/fibre_server.go
rootulp: Got it. I can remove the isValidator check and allow non-validators to start a Fibre server.

cc: @Wondertan because this diverges from what we originally discussed.

cmd/celestia-appd/cmd/start_command_standalone.go
rootulp: I couldn't find an easy way to extract this function

cmd/celestia-appd/cmd/fibre_server.go
rootulp: f276a9df1aa146b730c64c6285e6db86982f1c11

rootulp: On second thought I don't think this makes sense. Fibre server accepts as a first argument a privVal and non-validators won't have this.

Ref: 
- https://github.com/rootulp/celestia-app-fibre/blob/aa0e6d975ea9d23ea103ffd0b49595d38a9b507f/fibre/server.go#L64-L65

rootulp: https://github.com/celestiaorg/celestia-app-fibre/pull/59/commits/f7e0dd4778d4e1b63fa888b260824dc1fb69a87f

cmd/celestia-appd/cmd/fibre_server.go
rootulp: Created a FLUP: https://github.com/celestiaorg/celestia-app-fibre/issues/61

cmd/celestia-appd/cmd/root.go
Wondertan: "Why the block time is actually needed" question aside, this config field was not meant to be exposed to be configured by user, especially as cli flag. 

The intention is that the code which sets up the server picks the right value based on consensus timeout params automatically. This is similar to chainid param which should not have configuration diverging from apps.

cmd/celestia-appd/cmd/fibre_server_standalone.go
```diff
@@ -0,0 +1,38 @@
...
+	return fibre.SetupServer(fibre.ServerSetupConfig{
+		Node:       cmtNode,
+		GRPCServer: grpcServer,
+		GRPCClient: clientCtx.GRPCClient,
+		Logger:     svrCtx.Logger,
+		RootDir:    svrCtx.Config.RootDir,
+		Enabled:    svrCtx.Viper.GetBool(FibreEnableFlag),
+		StoreType:  svrCtx.Viper.GetString(FibreStoreTypeFlag),
+		StorePath:  svrCtx.Viper.GetString(FibreStorePathFlag),
+		ChainID:    svrCtx.Viper.GetString("chain-id"),
```
Wondertan: Also extract as a constant like others? 

Why would we actually need to configure it for fibre specifically? The app should know which chain id it runs with its own configuration and we shouldnt allow fibre to have a different chain id configured

fibre/server_setup.go
```diff
@@ -0,0 +1,133 @@
...
+	// StoreTypeMemory represents an in-memory store (ephemeral, non-persistent)
+	StoreTypeMemory = "memory"
+	// StoreTypeBadger represents a BadgerDB store (persistent on disk)
+	StoreTypeBadger = "badger"
+)
+
+// Logger interface abstracts logging functionality
+type Logger interface {
+	Info(msg string, keyvals ...interface{})
+}
```
Wondertan: Why do we need to abstract the logger? 

Also, server expects an std logger, but we are not populating it with this logger given in setup

fibre/server_setup.go
```diff
@@ -0,0 +1,133 @@
...
+	fibregrpc "github.com/celestiaorg/celestia-app/v6/fibre/grpc"
+	"github.com/celestiaorg/celestia-app/v6/x/fibre/types"
+	"github.com/cometbft/cometbft/node"
+	coregrpc "github.com/cometbft/cometbft/rpc/grpc"
+	"google.golang.org/grpc"
+)
+
+const (
+	// StoreTypeMemory represents an in-memory store (ephemeral, non-persistent)
+	StoreTypeMemory = "memory"
```
Wondertan: The memory store exists for tests only and is not intended to be used for anything else. I would remove storage type selection from user and for now hardcode badger only.

fibre/server_setup.go
```diff
@@ -0,0 +1,133 @@
...
+	// Logger is used for logging
+	Logger Logger
+	// RootDir is the root directory for default store paths
+	RootDir string
+	// Enabled indicates whether the Fibre server should be started
+	Enabled bool
+	// StoreType is the store type: StoreTypeMemory or StoreTypeBadger
+	StoreType string
+	// StorePath is the path for the badger store (only used if StoreType is StoreTypeBadger)
+	StorePath string
```
Wondertan: This is not needed. Badger by default creates additonal folder under the root path given to it. Also, there is no point in configuring this for the user.

fibre/server_setup.go
```diff
@@ -0,0 +1,133 @@
...
+
+// SetupServer initializes and registers the Fibre server with the gRPC server for a validator node.
+// Returns the Fibre server instance and an error. The server should be stopped gracefully during shutdown.
+// If the node is not a validator (no usable PrivValidator), this function does nothing and returns nil, nil.
+// If the node is a validator and initialization fails, returns an error (preventing node startup).
+func SetupServer(config ServerSetupConfig) (*Server, error) {
+	if !config.Enabled {
+		config.Logger.Info("Fibre server is disabled via flag, skipping Fibre server startup")
+		return nil, nil
+	}
```
Wondertan: I really wanna go full linus mode saying that the code like this actually makes the world a worse place. We dont need to pull flag specific logging into a fibre pkg and even making a field on the weird secondary struct just to make this log

fibre/server_setup.go
```diff
@@ -0,0 +1,133 @@
...
+)
+
+// Logger interface abstracts logging functionality
+type Logger interface {
+	Info(msg string, keyvals ...interface{})
+}
+
+// ServerSetupConfig contains all the dependencies needed to set up a Fibre server.
+// This allows sharing the setup logic between multiplexer and non-multiplexer builds.
+type ServerSetupConfig struct {
```
Wondertan: After all the comments above are resolved, this struct can be dissolved, leaving the remaining as normal constructor params

Generally,  there is already a server config struct in this pkg and we should stick to using.

Wondertan (dismissed):

cmd/celestia-appd/cmd/root.go
rootulp: Ack will remove block-time flag.

cmd/celestia-appd/cmd/fibre_server_standalone.go
rootulp: > Why would we actually need to configure it for fibre specifically?

We need to provide the chain-id to the Fibre server because this default is wrong: https://github.com/rootulp/celestia-app-fibre/blob/dda61fcfcabb11f9513430d514ee7be11134fa7e/fibre/server.go#L37

On a related note https://github.com/rootulp/celestia-app-fibre/blob/dda61fcfcabb11f9513430d514ee7be11134fa7e/fibre/server.go#L37 should've used an extracted constant.

I'll extract a constant for chain-id.

fibre/server_setup.go
rootulp: That wasn't obvious to me but can do.

fibre/server_setup.go
rootulp: > I really wanna go full linus mode saying that the code like this actually makes the world a worse place.

This feedback isn't actionable and it's pretty rude. Please try to make your feedback clear and actionable.

> We dont need to pull flag specific logging into a fibre pkg and even making a field on the weird secondary struct just to make this log

I don't understand the issue or your proposed solution.

cmd/celestia-appd/cmd/root.go
rootulp: d81471fbda58eb5d824eafa9985674f88307160e

cmd/celestia-appd/cmd/fibre_server_standalone.go
rootulp: 05761a0179bdddecca092f504317fbc9ce186772

fibre/server_setup.go
rootulp: Removed logger interface: 08d729c48475d56797fa04f854fa77ef7c5d4665

fibre/server_setup.go
rootulp: Removed store type memory and hard-coded badger: 4a7f2fa0fb7687bb2ba8203abe6e4d5a86b3fc50

fibre/server_setup.go
rootulp: Removed 20760552d0974fb342dff154a4b62aa426b7ede3

fibre/server_setup.go
rootulp: I removed this code block because over Slack you said that the Fibre enable flag should be removed. I'm not sure what else you take issue with so I'm marking this as resolved.

fbfb96bc8e8ad75837b73239ad5094d4c7c4d97f

Wondertan: > This feedback isn't actionable and it's pretty rude. Please try to make your feedback clear and actionable.

My apologies, I didn't mean to offend you, nor did I think you would take this personally, since it looked like AI slop. This is a reference to a recent [code review from linus](https://www.tomshardware.com/software/linux/linus-torvalds-calls-risc-v-code-from-google-engineer-garbage-and-that-it-makes-the-world-actively-a-worse-place-to-live-linux-honcho-puts-dev-on-notice-for-late-submissions-too), which I thought would be funny.

cmd/celestia-appd/cmd/fibre_server.go
cmwaters: Just so we're on the same page: a validator is only a validator when they enter the set. We can have people who are outside the set and are running a consensus node who should still be running the fibre server, it just won't be used.

cmd/celestia-appd/cmd/root.go
cmwaters: Yeah I think we should avoid assuming block time here if we can.

fibre/server_setup.go
Wondertan: > I don't understand the issue or your proposed solution.

The problem is that this if statement brought zero value, while it added an additional field to this config structure(which we shouldn't need either btw) and spread the logic of enabling server across multiple pkgs. It also pulls in the context of flags and cli into the core `fibre` pkg for no good reason. 

As to proposed solutions, this simply could be an if statement in the cli logic that doesn't call the fibre constructor if not enabled.

Wondertan: It's also an anti-pattern to return double nils.

rootulp: Thanks that helps clear some things up. Is https://github.com/celestiaorg/celestia-app-fibre/commit/fbfb96bc8e8ad75837b73239ad5094d4c7c4d97f sufficient? It removes the conditional, and the enable flag.

cmd/celestia-appd/cmd/fibre_server_standalone.go
Wondertan: > We need to provide the chain-id to the Fibre server because this default is wrong: https://github.com/rootulp/celestia-app-fibre/blob/dda61fcfcabb11f9513430d514ee7be11134fa7e/fibre/server.go#L37

Ok, sure, but it doesn't mean the user should be controlling the chain-id. The override can happen using the chain-id already in app's context. Allowing the user to change it is a footgun leading to server and app's chain-ids conflicting

cmd/celestia-appd/cmd/root.go
Wondertan: But can't we get the actual value instead of relying on arbitrary default?

fibre/server_setup.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/cometbft/cometbft/node"
+	coregrpc "github.com/cometbft/cometbft/rpc/grpc"
+	"google.golang.org/grpc"
+)
+
+// ServerSetupConfig contains all the dependencies needed to set up a Fibre server.
+// This allows sharing the setup logic between multiplexer and non-multiplexer builds.
+type ServerSetupConfig struct {
+	// Node is the CometBFT node instance
+	Node *node.Node
```
Wondertan: Why pull the entire node if we only need privval?

cmd/celestia-appd/cmd/fibre_server_standalone.go
rootulp: Yes, that's what https://github.com/celestiaorg/celestia-app-fibre/commit/05761a0179bdddecca092f504317fbc9ce186772 does.

cmd/celestia-appd/cmd/root.go
rootulp: There is no actual value for block time. Celestia consensus targets a block time (e.g. 6 seconds) but the actual block time can deviate from it.

Wondertan: > block time can deviate from it. 

This is fine for what it is used for atm. It is the same for using the default.

> There is no actual value for block time.

Isn't there a consensus timeout params that controls block time, and couldn't we derive block time from them? My assumption is that it shouldn't be hard, but if there is some hassle in doing this, I am fine with defaults. We just may forget bumping it once we decide to decrease it, lets say in v8

Wondertan: There is also a chance we remove block time ofc

rootulp: 6ec931b4c114c34e188192d82b895061e38abfbe

fibre/server_setup.go
rootulp: TODO: only propagate PrivVal.

fibre/server_setup.go
rootulp: 4cbb092fcc23080404fdc42de04113933bfd39e0

fibre/server_setup.go
rootulp: 14301d8ef7879a5233be63782205efc680f60d99

rootulp:
TODO: remove `SetupServer` and collapse it into `NewServer`

cmd/celestia-appd/cmd/start_command_standalone.go
```diff
@@ -0,0 +1,339 @@
...
+		grpcServer, clientCtx, err = createGRPCServer(svrCtx, clientCtx, appInstance, svrCfg, cmtNode)
+		if err != nil {
+			return fmt.Errorf("failed to create gRPC server: %w", err)
+		}
+
+		// Register Fibre server BEFORE starting the gRPC server
+		serverConfig := fibre.DefaultServerConfig()
+		serverConfig.ChainID = svrCtx.Viper.GetString(ChainIDKey)
+		serverConfig.StoreConfig.Path = filepath.Join(svrCtx.Config.RootDir, "data", "fibre-store")
+		// TODO: convert the svrCtx.Logger into a *slog.Logger and then propgate
```
rootulp, line 107: https://github.com/celestiaorg/celestia-app-fibre/issues/62

multiplexer/abci/multiplexer.go
```diff
@@ -173,12 +174,47 @@ func (m *Multiplexer) enableGRPCAndAPIServers(app servertypes.Application) error
...
 		m.clientContext = clientContext // update client context with grpc
 
+		// Register Fibre server BEFORE starting the gRPC server
+		// This ensures all services are registered before Server.Serve() is called
+		var fibreServer *fibre.Server
+		if m.cmNode != nil {
+			serverConfig := fibre.DefaultServerConfig()
+			serverConfig.ChainID = m.chainID
+			serverConfig.StoreConfig.Path = filepath.Join(m.svrCtx.Config.RootDir, "data", "fibre-store")
+			// TODO: convert the m.Logger into a *slog.Logger and then propgate
```
rootulp, line 191: https://github.com/celestiaorg/celestia-app-fibre/issues/62

cmd/celestia-appd/cmd/start_command_standalone.go
cmwaters: Ah does this have to have these arguments to fit some function signature?

evan-forbes (dismissed):
late to review, but overall seems pretty straightforward. lgtm!

cmd/celestia-appd/cmd/root.go
```diff
@@ -39,6 +39,9 @@ const (
 
 	// DelayedPrecommitTimeoutFlag is a flag that can be used to override the DelayedPrecommitTimeout.
 	DelayedPrecommitTimeoutFlag = "delayed-precommit-timeout"
+
+	// ChainIDKey is the viper key for the chain ID
+	ChainIDKey = "chain-id"
```
cmwaters, line 44: more of a FLUP as it's not necessary now - but it would be nice to not require the chain-id in setup. This is just to minimise the flags etc. that the user has to provide to setup the fibre server

cmd/celestia-appd/cmd/start_command_standalone.go
```diff
@@ -0,0 +1,339 @@
...
+	cfg := svrCtx.Config
+
+	// Load or generate node key
+	nodeKey, err := p2p.LoadOrGenNodeKey(cfg.NodeKeyFile())
+	if err != nil {
+		return nil, fmt.Errorf("failed to load or generate node key: %w", err)
+	}
+
+	// Load or generate private validator
+	privVal := pvm.LoadOrGenFilePV(cfg.PrivValidatorKeyFile(), cfg.PrivValidatorStateFile())
```
cmwaters, line 161: [not blocking] Does this support KMS?

cmd/celestia-appd/cmd/start_command_standalone.go
```diff
@@ -0,0 +1,339 @@
...
+	_, _, err := net.SplitHostPort(svrCfg.GRPC.Address)
+	if err != nil {
+		return nil, clientCtx, fmt.Errorf("invalid gRPC address: %w", err)
+	}
+
+	// Create gRPC client for gateway
+	grpcClient, err := grpc.NewClient(
+		svrCfg.GRPC.Address,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	)
```
cmwaters, line 210: We might need to add an option to configure message sizes. I think by default this is only 4KB

fibre/server.go
```diff
@@ -59,15 +62,21 @@ type Server struct {
...
 // Returns an error if the validator's public key cannot be retrieved.
 func NewServer(
 	privVal core.PrivValidator,
-	queryClient types.QueryClient,
-	valGet validator.SetGetter,
-	store *Store,
 	cfg ServerConfig,
+	grpcServer *grpc.Server,
+	grpcClient *grpc.ClientConn,
 ) (*Server, error) {
```
cmwaters, line 72: Would it make sense to have this as a separate method like `NewGRPCServer` because if I'm not mistaken, requiring the grpc server as an argument makes it difficult to use other implementations of the interfaces for mocking and for the light/bridge node

cmwaters (commented):
Looks good, only have two comments of relevance

cmd/celestia-appd/cmd/start_command_standalone.go
Wondertan: This is not GRPC client used to query the fibre server, but query SDK module in the same process and there 4KB should be sufficient.

fibre/server.go
Wondertan: This is the change I suggested to Rootul, but I think you have a point here. We won't be able to easily mock the GRPC server if we pass in concrete values. Having an additional constructor would solve this.

cmd/celestia-appd/cmd/start_command_standalone.go
Wondertan: Rootul and I checked the implementation of the cometbft node, and according to the code it should work. The node constructor force overrides this filePV with the KMS one if configured. This is dirty and confusing from an API design perspective, but it will work.

cmd/celestia-appd/cmd/root.go
Wondertan: https://github.com/celestiaorg/celestia-app-fibre/pull/59#discussion_r2492887111

I was similarly confused there. This chainID is actually just a constant for the flag in Viper that already exists, and this constant just defines it as a constant rather than being used as a magic string.

cmd/celestia-appd/cmd/start_command_standalone.go
rootulp: Pretty sure it does by created https://github.com/celestiaorg/celestia-app-fibre/issues/61 to actually try it

cmd/celestia-appd/cmd/root.go
rootulp: Agreed. chainId is no longer required as a flag, the Fibre server gets set up with whatever chain-id was configured in the server context.

fibre/server.go
rootulp: fd110e61f6835bcb23efbaaca7c9241fce6c3836

cmd/celestia-appd/cmd/start_command_standalone.go
rootulp: I assume this is handled by https://github.com/celestiaorg/celestia-app-fibre/pull/70 so marking this as resolved.

evan-forbes (approved):

PR #64, opened 2025-11-05
fix: register fibre module account

cmwaters:
The fibre module account was not registered in the maccPerms map.

```
Error Trace:    /Users/callum/Developer/go/src/github.com/celestiaorg/celestia-app-fibre/fibre/fibre_e2e_test.go:269                            
                Error:          Received unexpected error:
                                tx execution failed with code 111222: recovered: module account fibre does not exist: unknown address [celestiaorg/cosmos-sdk@v0.51.3/x/bank/keeper/keeper.go:299]  
```

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -144,6 +144,7 @@ var maccPerms = map[string][]string{
 	icatypes.ModuleName:            nil,
 	hyperlanetypes.ModuleName:      nil,
 	warptypes.ModuleName:           {authtypes.Minter, authtypes.Burner},
+	fibretypes.ModuleName:          nil,
 }
 
 var (


rootulp (approved):
LGTM. Traced the call stack and it looks like the bank keeper -> auth keeper -> GetModuleAccount -> GetModuleAccountAndPermissions -> GetModuleAddressAndPermissions -> which looks up the address in a map of permissions

https://github.com/rootulp/cosmos-sdk/blob/058b4b769b728d26705108087b59415d95ba3161/x/auth/keeper/keeper.go#L91

evan-forbes (approved):

PR #66, opened 2025-11-05
fix: go lint issues

rootulp:
Fixes a bunch of golangci-lint issues. Not sure how/why these merged to feature/fibre.

Update: looks like CI isn't running checks after PRs merge to feature/fibre. And they're not run on this PR.

## Testing

```
make lint
# passes go lint
```

There are a bunch of markdownlint failures but those seem present on celestia-app too.


cmwaters (approved):
utack

PR #68, opened 2025-11-06
fix(fibre/validator): secure assignment algorithm

Wondertan:
Implements the algorithm described in https://github.com/celestiaorg/celestia-app/issues/6670#issuecomment-3485369784. Besides, we use chacha8 instead of chacha20, which was approved by @nashqueue.

Pair-programmed with @nashqueue.

~Todo: update benchmark numbers.~

### DIFF
--- a/fibre/validator/set.go
+++ b/fibre/validator/set.go
@@ -2,7 +2,7 @@ package validator
 
 import (
 	"fmt"
-	"math/big"
+	"math/rand/v2"
 
 	"github.com/celestiaorg/rsema1d"
 	"github.com/cometbft/cometbft/crypto"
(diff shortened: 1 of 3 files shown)


fibre/validator/set.go
```diff
@@ -31,33 +31,42 @@ func (s Set) GetByAddress(address crypto.Address) (*core.Validator, bool) {
...
+	rng.Shuffle(totalRows, func(i, j int) {
+		rowsIndicies[i], rowsIndicies[j] = rowsIndicies[j], rowsIndicies[i]
+	})
+
+	// assign rows to validators in a ShardMap
 	shardMap := make(ShardMap)
+	for i, validator := range s.Validators {
+
+		// TODO(@Wondertan): As per Nashqueue, we no longer want to send every row to validators and some might be not assigned.
+		// So the number of rows to assign should be given as parameter to Assign and which probably would be taken from BlobConfig.
```
Wondertan, line 63: Note to myself to make an issue for this

Wondertan: or may even do in this PR

Wondertan:
As per [d9304e9](https://github.com/celestiaorg/celestia-app-fibre/pull/68/commits/d9304e90d030b3a3c21ae1b54ce7a45a855a2dbf), we can see that the new algorithm is ~14x faster (1.47ms → 105µs)  and uses ~23x less memory  (2.87MB → 132KB) than the previous mock algorithm, essentially due to `big.Int` overhead (although it worth mentioning that the previous algo didn't use proper uint256 implementation)

walldiss (approved):
lgtm

fibre/validator/set.go
walldiss: So we miss only those that do not get assigned due to int division? In other works each val gets same amount of rows and remainder is just not sent?

Wondertan: Wdym "not get assigned due to int division"

Wondertan: https://github.com/celestiaorg/celestia-app-fibre/issues/80

PR #70, opened 2025-11-06
fix(fibre): grpc support for large blobs

### LINKED ISSUE #65: grpc message sizes should match maximum blob size
In an e2e test, I got the following error when trying to submit a 128KB blob. 

```
2025/11/05 16:08:25 WARN failed to upload rows component=fibre-client validator=8536FD49908FEB5DDF828DFA548B9CBBEB48B9C2 blob_commitment=ebbe31bcc2e006c8c128f31be25b0491c9507f1593f2ac45caa2099040bb3f1c rows_count=16384 error="rpc error: code = ResourceExhausted desc = grpc: received message larger than max (9044046 vs. 4194304)" 
```

The `DefaultNewClientFn` should set the grpc message size limits in mind of the 128MB upper bound.

Wondertan:
Make sure that both GRPC client and server configurations support sending and receiving the largest blob size. It dynamically calculates the max size based on config values, including ShardingFactor(number of validators). 

Tested with integration testing that runs real grpc client and server.

Closes #65

### DIFF
--- a/fibre/blob.go
+++ b/fibre/blob.go
@@ -1,6 +1,7 @@
 package fibre
 
 import (
+	"crypto/sha256"
 	"encoding/binary"
 	"encoding/hex"
 	"errors"
@@ -52,17 +53,22 @@ type BlobConfig struct {
 	BlobVersion uint8
 	// CodingWorkers is the number of workers to use for encoding and decoding rsema1d.
 	CodingWorkers int
+	// ShardingFactor is the expected number of validators in the network.
+	// This is used to calculate the maximum gRPC message size per validator.
+	// Should be set to match the expected validator set size.
+	ShardingFactor int
 }
 
 // DefaultBlobConfigV0 returns a [BlobConfig] with default values for version 0.
 func DefaultBlobConfigV0() BlobConfig {
 	return BlobConfig{
-		OriginalRows:  4096,
-		ParityRows:    12288, // (3 * OriginalRows, TotalRows = 16384)
-		RowSizeMin:    64,
-		MaxBlobSize:   128 * 1024 * 1024,
-		BlobVersion:   0,
-		CodingWorkers: runtime.GOMAXPROCS(0),
+		OriginalRows:   4096,
+		ParityRows:     12288, // (3 * OriginalRows, TotalRows = 16384)
+		RowSizeMin:     64,
+		MaxBlobSize:    128 * 1024 * 1024,
+		BlobVersion:    0,
+		CodingWorkers:  runtime.GOMAXPROCS(0),
+		ShardingFactor: 100, // Expected number of validators
 	}
 }
 
@@ -95,6 +101,33 @@ func (c BlobConfig) MaxRowSize() int {
 // This is the size included in the [PaymentPromise] and the one actually paid for.
 func (c BlobConfig) UploadSize(dataLen int) int {
 	return c.RowSize(dataLen) * c.OriginalRows
+}
+
+// MaxShardSize calculates the maximum size of a shard (subset of blob rows assigned to a validator with RLC and Merkle proofs)
+// This does not include protocol overhead like PaymentPromise or protobuf encoding overhead.
+func (c BlobConfig) MaxShardSize() int {
+	const (
+		rowIndexSize = 4  // uint32 index per row
+		rlcCoeffSize = 16 // uint128 coefficient per row
+	)
+
+	totalRows := c.OriginalRows + c.ParityRows
+	maxRowSize := c.MaxRowSize()
+	rlcCoeffsSize := c.OriginalRows * rlcCoeffSize
+
+	// get proof size per row by finding merkle tree depth
+	treeDepth := 0
+	for n := totalRows; n > 1; n = (n + 1) / 2 {
+		treeDepth++
+	}
+	proofSizePerRow := treeDepth * sha256.Size
+
+	// calculate rows per validator based on sharding factor
+	// add 1 to account for potential uneven distribution
+	// TODO(@Wondertan): This is not completely accurate, but it's a good approximation for now.
+	rowsPerValidator := (totalRows / c.ShardingFactor) + 1
+
+	return rlcCoeffsSize + (rowsPerValidator * (rowIndexSize + maxRowSize + proofSizePerRow))
 }
 
 // Blob represents encoded data with Reed-Solomon erasure coding.
--- a/fibre/client.go
+++ b/fibre/client.go
@@ -109,7 +109,7 @@ func NewClient(txClient *user.TxClient, kr keyring.Keyring, valGet validator.Set
 	}
 
 	if cfg.NewClientFn == nil {
-		cfg.NewClientFn = grpc.DefaultNewClientFn(hostReg)
+		cfg.NewClientFn = grpc.DefaultNewClientFn(hostReg, MaxMessageSize(cfg.BlobConfig))
 	}
 	if cfg.Tracer == nil {
 		cfg.Tracer = otel.Tracer("fibre-client")
@@ -152,4 +152,10 @@ func (c *Client) Close() error {
 
 	c.closeWg.Wait()
 	return c.clientCache.Close()
+}
+
+// MaxMessageSize returns the maximum message size that can be sent over the network.
+func MaxMessageSize(cfg BlobConfig) int {
+	msgSize := cfg.MaxShardSize() + MaxPaymentPromiseSize
+	return msgSize + (msgSize / 50) // add 2% protobuf overhead
 }
--- a/fibre/payment_promise.go
+++ b/fibre/payment_promise.go
@@ -123,9 +123,12 @@ func (p *PaymentPromise) Validate() error {
 		return fmt.Errorf("signer key must be %d bytes, got %d", secp256k1.PubKeySize, len(p.SignerKey.Key))
 	}
 
-	// chain ID must not be empty
+	// chain ID must not be empty and within length limit
 	if p.ChainID == "" {
 		return errors.New("chain id must not be empty")
+	}
+	if len(p.ChainID) > maxChainIDSize {
+		return fmt.Errorf("chain id length %d exceeds maximum %d", len(p.ChainID), maxChainIDSize)
 	}
 
 	// upload size must be positive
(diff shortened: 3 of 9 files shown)


fibre/blob.go
```diff
@@ -52,17 +53,22 @@ type BlobConfig struct {
 	BlobVersion uint8
 	// CodingWorkers is the number of workers to use for encoding and decoding rsema1d.
 	CodingWorkers int
+	// ShardingFactor is the expected number of validators in the network.
+	// This is used to calculate the maximum gRPC message size per validator.
+	// Should be set to match the expected validator set size.
+	ShardingFactor int
```
rootulp, line 59: [nit] `ShardingFactor` is indirect and isn't as descriptive as something like `NumValidators` or `ValidatorCount`

```suggestion
	// ValidatorCount is the expected number of validators in the network.
	// This is used to calculate the maximum gRPC message size per validator.
	// Should be set to match the expected validator set size.
	ValidatorCount int
```

also this should probably say number of active validators because there are 269 validators on Mainnet but only 100 are in the active set. More specifically it could say the number of Fibre Service Providers.

Ref: https://www.mintscan.io/celestia/validators

fibre/blob.go
```diff
@@ -97,6 +103,33 @@ func (c BlobConfig) UploadSize(dataLen int) int {
...
+	}
+	proofSizePerRow := treeDepth * sha256.Size
+
+	// calculate rows per validator based on sharding factor
+	// add 1 to account for potential uneven distribution
+	// TODO(@Wondertan): This is not completely accurate, but it's a good approximation for now.
+	rowsPerValidator := (totalRows / c.ShardingFactor) + 1
+
+	return rlcCoeffsSize + (rowsPerValidator * (rowIndexSize + maxRowSize + proofSizePerRow))
+}
```
rootulp, line 131: We can delete all of this code if we just set the max message size to a fixed constant.

fibre/client.go
```diff
@@ -153,3 +153,9 @@ func (c *Client) Close() error {
 	c.closeWg.Wait()
 	return c.clientCache.Close()
 }
+
+// MaxMessageSize returns the maximum message size that can be sent over the network.
+func MaxMessageSize(cfg BlobConfig) int {
+	msgSize := cfg.MaxShardSize() + MaxPaymentPromiseSize
+	return msgSize + (msgSize / 50) // add 2% protobuf overhead
+}
```
rootulp, line 161: If this calculation is wrong, clients will still hit https://github.com/celestiaorg/celestia-app-fibre/issues/65. IMO this function could be deleted and we can set the max message size to some constant. For example: `math.MaxInt32`

Ref:

- https://github.com/rootulp/celestia-app-fibre/blob/fd110e61f6835bcb23efbaaca7c9241fce6c3836/multiplexer/abci/multiplexer.go#L316-L317

fibre/payment_promise.go
```diff
@@ -167,12 +170,23 @@ func (p *PaymentPromise) Validate() error {
 }
 
 const (
+	// MaxPaymentPromiseSize is the theoretical minimum size of all PaymentPromise fields
+	// before protobuf encoding overhead
+	MaxPaymentPromiseSize = signBytesFixedSize + signatureSize + maxChainIDSize
```
rootulp: [blocking] I don't follow. How can this be the max size but also the minimum size?

fibre/payment_promise.go
```diff
@@ -167,12 +170,23 @@ func (p *PaymentPromise) Validate() error {
...
 	signBytesPrefix = "fibre/pp:v0"
 	// signBytesFixedSize is the size of all the constant fixed size fields.
-	// Format: signerPubKey(33) + namespace(29) + blobSize(4) + commitment(32) + blobVersion(4) + height(8)
-	signBytesFixedSize = secp256k1.PubKeySize + share.NamespaceSize + 4 + 32 + 4 + 8
+	// Format: signerPubKey(33) + namespace(29) + blobSize(4) + commitment(32) + blobVersion(4) + height(8) + timestamp(15)
+	signBytesFixedSize = secp256k1.PubKeySize + share.NamespaceSize + 4 + 32 + 4 + 8 + 15
+
+	// maxChainIDSize is the maximum allowed chain ID length.
+	// Examples: "celestia" (8), "mocha-4" (7), "arabica-11" (10)
+	maxChainIDSize = 20
```
rootulp, line 186: [blocking] why add this? Is there a max chain-id length enforced anywhere else in the codebase? If not, let's remove

fibre/payment_promise.go
```diff
@@ -192,9 +206,11 @@ func (p *PaymentPromise) SignBytes() ([]byte, error) {
 			p.signBytesErr = fmt.Errorf("marshalling timestamp: %w", err)
 			return
 		}
+		// trim the last zone byte which is always UTC
+		timestampBytes = timestampBytes[:len(timestampBytes)-1]
 
 		// calculate total size including the prefix
-		totalSize := len(signBytesPrefix) + len(p.ChainID) + signBytesFixedSize + len(timestampBytes)
+		totalSize := len(signBytesPrefix) + len(p.ChainID) + signBytesFixedSize
```
rootulp: [question] why did these change? They don't seem related to a GRPC message size fix.

rootulp (changes requested):

fibre/client.go
Wondertan: It's not wrong, and the new test confirms this. It is insecure to set the max uint there, as it's a DOS issue.  It should use this function to get the real max message size, based on actual number of validators.

fibre/payment_promise.go
Wondertan: The comment is wrong; I will fix it.

fibre/payment_promise.go
Wondertan: I don't think there is a limit on the max chain ID length anywhere, while there should be one

fibre/blob.go
Wondertan: The higher-level constants can be descriptive on the protocol level. Ideally, there should be a canonical configuration, somehow like in `consts` that takes proper values named according to the core protocol. For the BlobConfig level, the validators are an external concept

fibre/blob.go
Wondertan: No, we can't. It should be dynamic to prevent overallocation of the allowed message size; otherwise, it's a DOS vector.

fibre/payment_promise.go
rootulp: Why?

fibre/payment_promise.go
Wondertan: That's true, that is quite unrelated. I can extract this in a separate PR if you insist.

fibre/client.go
rootulp: Do you think that the unit tests are exhaustive?

I don't see a test case for uploading MaxBlobSize to a single validator which naively seems like the worst case.

Wondertan: There is a test with 3 validators, though. I can change the number in the test to 1 if you want, but the overhead is a constant percentage, and it works with number being set to 1 as well

fibre/payment_promise.go
Wondertan: * to have a fixed max payment promise length 
* strings are unbounded, and we should bound everything, as otherwise it's a dos vector too.

fibre/blob.go
Wondertan: Needing to differentiate between active and non-active validators is also a reason not to bring in the notion of validators here. This parameter controls how many shards to split the blob into, not how many validators there is.

fibre/client.go
Wondertan: I changed it just in case, but again, the encoding overhead percentage is constant; it doesn't scale with the sharding factor

fibre/payment_promise.go
Wondertan: I removed the first line as it is indeed unrelated. The changes in the constants are related to this change though, so I am keeping them

walldiss:
Why are we worried about msg size limitation on client side? Are we trying to mitigate the case when validator will send too large messages to clients? It seems very unlikely and not a viable attack. I would just put some arbitrary large value there or remove the limitation of incoming and outgoing msg size on client side.

Wondertan:
> Are we trying to mitigate the case when validator will send too large messages to clients?

It's the opposite too. The client sends too big a value to the server, which is a real attack

walldiss:
> It's the opposite too. The client sends too big a value to the server, which is a real attack

Client who wants to send too large packets will modify your code. For this particular reason limits needs to be set on server side, which is not implemented in scope of this PR per description.

Wondertan:
>  For this particular reason limits needs to be set on server side, which is not implemented in scope of this PR per description.

The fact that the server configuration wasn't changed in this PR doesn't change the need for accurate message calculation for server configuration to happen. And again, as per the PR description, it wasn't changed because the grpc server is out of the fibre's control(and is shared with app), but this is exactly the reason why `MaxMessageSize` is exposed.

walldiss:
Thanks for the clarification—sounds like we’re discussing two different threat models:

1. Server → Client (oversized responses): I don’t think this is a viable attack in our context. For this case, we can either set a very high client limit or remove the client-side cap entirely to keep things simple.

2. Client → Server (oversized requests): This is a real attack, but client-side checks won’t stop a malicious client from sending large payloads. Mitigation must be enforced on the server (e.g., gRPC receive/send limits). As you noted, that configuration lives outside the scope of the fibre server and this PR.

Given that, I suggest we:
- Keep the client logic minimal (generous default or no limit).
- Clearly document that server-side limits are required to address (2) and will be handled in a separate change outside fibre.

Wondertan:
Given that we need to calculate value for the server and server with client reuse the same code, I believe we can just use the server value on the client too without the need to come up with an additional value.

As to server configuration, given that the server wiring to app was merged since this PR was opened I may actually include the change here too, should be like one liner. Will need to see how to do that first. Happy to do it in a separate PR too tho

fibre/payment_promise.go
Wondertan: fixed

Wondertan:
Added server-side configuration so we don't have any follow-ups here and close the issue entirely, should be good for rereview now.

fibre/blob.go
rootulp: `ShardingFactor` forces readers to learn a new term and understand a new relationship (`ShardingFactor` == number of Fibre service providers).

IMO using a new name for this adds cognitive complexity. We expect this parameter to  match the number of Fibre service providers so I still vote for a name like that but not blocking on it

fibre/blob.go
rootulp: I still think it's way easier to understand and maintain this code if we used a fixed constant.

fibre/payment_promise.go
rootulp: I'm concerned this will break Fibre for a subset of development chains because a chain-id max length is not enforced elsewhere.

If you think chain-id should be bounded then IMO that warrants a FLUP issue and a corresponding PR to cap it in multiple places (not just in Fibre).

rootulp (approved):

PR #77, opened 2025-11-08
feat: route MsgPayForFibre transactions to PayForFibreNamespace

rootulp:
Blocked on https://github.com/celestiaorg/go-square/pull/202

### DIFF
--- a/app/filtered_square_builder.go
+++ b/app/filtered_square_builder.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"github.com/celestiaorg/celestia-app/v6/pkg/appconsts"
+	fibretypes "github.com/celestiaorg/celestia-app/v6/x/fibre/types"
 	square "github.com/celestiaorg/go-square/v3"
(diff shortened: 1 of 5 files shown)


app/filtered_square_builder.go
```diff
@@ -190,8 +238,31 @@ func separateTxs(_ client.TxConfig, rawTxs [][]byte) ([][]byte, []*tx.BlobTx) {
...
+				normalTxs = append(normalTxs, rawTx)
+				continue
+			}
+
+			if _, hasPayForFibre := extractMsgPayForFibre(sdkTx); hasPayForFibre {
+				payForFibreTxs = append(payForFibreTxs, rawTx)
+			} else {
+				normalTxs = append(normalTxs, rawTx)
+			}
+		}
```
evan-forbes, line 253: very optional but we might be able to clean up the if else nests here by 1 level by continue -ing after appending the blobTx on line 239

same goes for 249

evan-forbes (approved):

Issue #69, opened 2025-11-06
Make CI pass


rootulp:
## Problem

There are a few issues with CI in this repo

1. CI isn't run after PRs merge to `feature/fibre`
2. CI is failing on most PRs

rootulp:
🤞 this passes after #88 merges but will keep it open until that merges and I verify

rootulp:
Un-assigning myself b/c I'm not working on this in the short term.

rootulp:
No longer needed b/c this repo is about to be archived.

PR #88, opened 2025-11-11
ci: attempt to fix build in CI

rootulp:
Part of #69 

Attempts to fix Build failing in CI on PRs. [Example](https://github.com/celestiaorg/celestia-app-fibre/actions/runs/19234895597/job/54982107615?pr=79)


liamsi (approved):
utACK 
(at least GOPRIVATE is what I use locally too)

PR #102, opened 2025-11-20
feat(app): support pay for fibre txs in ProcessProposal

### LINKED ISSUE #47: Write MsgPayForFibre to data square
## Context

Spun out of #1 

## Problem

The spec says when processing a MsgPayForFibre:
- The transaction should be included in the reserved namespace for Fibre transactions (PAY_FOR_FIBRE_NAMESPACE = 0x0000000000000000000000000000000000000000000000000000000005)
- A system-level blob should be generated with the namespace from PaymentPromise and blob data as the Fibre blob commitment

This hasn't yet been implemented. That means commitments aren't included in the data square.

## Proposal

- [x] https://github.com/celestiaorg/go-square/pull/200
- [x] https://github.com/celestiaorg/go-square/pull/201
- [x] https://github.com/celestiaorg/go-square/pull/202
- [x] https://github.com/celestiaorg/celestia-app-fibre/pull/77
- [x] In celestia-app-fibre generate a system-level blob for each MsgPayForFibre message
    - [x] https://github.com/celestiaorg/go-square/pull/205
    - [x] https://github.com/celestiaorg/celestia-app-fibre/pull/85
    - [x] https://github.com/celestiaorg/celestia-app-fibre/pull/102
- [x] https://github.com/celestiaorg/celestia-app-fibre/issues/104

rootulp:
Closes #47 

~Blocked on https://github.com/celestiaorg/go-square/pull/209~

Update ProcessProposal to handle PayForFibre txs.


app/process_proposal_test.go
```diff
@@ -0,0 +1,215 @@
...
+			name:          "invalid blob tx in middle - should report unmarshalling error",
+			txs:           [][]byte{normalTx, createMalformedBlobTxForTest(t), normalTx},
+			wantError:     true,
+			errorContains: "unmarshalling blob tx",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := validateTxOrdering(tt.txs, txConfig)
```
rootulp: [note] this is a new `process_proposal_test.go` file to test the private method validateTxOrdering

Although this could be extracted to go-square and tested there

rootulp:
Bridge nodes call `square.Construct` so if this logic doesn't exist in go-square then the bridge nodes can't reconstruct squares.

So we shouldn't merge this as-is. We need to do https://github.com/celestiaorg/go-square/issues/208

app/process_proposal.go
```diff
@@ -128,7 +130,16 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
 
 	}
 
-	eds, err := da.ConstructEDSWithTreePool(req.Txs, appconsts.Version, app.MaxEffectiveSquareSize(ctx), app.TreePool())
```
evan-forbes, line 131: this appears to only route on version so we probably could get rid of this function I think or just put it in the tool that needs to route based on version

I'll create a flup in app

app/filtered_square_builder.go
```diff
@@ -293,6 +300,39 @@ func separateTxs(txConfig client.TxConfig, rawTxs [][]byte) (normalTxs [][]byte,
...
+	sdkTx, err := h.txConfig.TxDecoder()(tx)
+	if err != nil {
+		return false
+	}
+	_, hasPayForFibre := extractMsgPayForFibre(sdkTx)
+	return hasPayForFibre
+}
+
+// CreateSystemBlob creates a system blob from a PayForFibre transaction.
+func (h *payForFibreHandler) CreateSystemBlob(tx []byte) (*share.Blob, error) {
```
evan-forbes, line 324: not blocking low prio question 

will other people using go-square also need to import this to construct a square? if so, is that an issue?

iirc some of the goals of go-square were so that we didn't have to import celestia-app to construct a square. if we don't need that property, then maybe we could simplify things somehow

evan-forbes (dismissed):

app/filtered_square_builder.go
rootulp: > will other people using go-square also need to import this to construct a square? if so, is that an issue?

Yes

> iirc some of the goals of go-square were so that we didn't have to import celestia-app to construct a square. if we don't need that property, then maybe we could simplify things somehow

The only consumers of go-square I can think of are: celestia-app, celestia-node. celestia-node imports go-square and celestia-app so I don't think we _need_ this property.

go.mod
```diff
@@ -282,7 +282,7 @@ require (
 replace (
 	cosmossdk.io/api => github.com/celestiaorg/cosmos-sdk/api v0.7.6
 	cosmossdk.io/x/upgrade => github.com/celestiaorg/cosmos-sdk/x/upgrade v0.2.0
-	github.com/celestiaorg/go-square/v3 => github.com/celestiaorg/go-square/v3 v3.0.0-20251112213826-ad7dfd52baef
+	github.com/celestiaorg/go-square/v3 => ../go-square
```
rootulp: I should create a go-square release from the feature/fibre branch and use that here before merging

rootulp: Created https://github.com/celestiaorg/go-square/releases/tag/v4.0.0-rc0

rootulp:
When I try to use go-square v4.0.0-rc1 I hit

```
panic: proto: file "proto/blob/v2/blob.proto" is already registered
		previously from: "github.com/celestiaorg/go-square/v3/proto/blob/v2"
		currently from:  "github.com/celestiaorg/go-square/v4/proto/blob/v2"
	See https://protobuf.dev/reference/go/faq#namespace-conflict
```

ahh I need to do something like https://github.com/celestiaorg/go-square/pull/194

evan-forbes (approved):

PR #108, opened 2025-11-21
fix(x/fibre): remove creation timestamp in future validation

### LINKED ISSUE #67: Using current timestamp is too far in the future
When running the e2e test, I've got the following error:

```
2025/11/05 16:12:27 WARN failed to upload rows component=fibre-client validator=BC62B2D8D70859E7289C2338E751B5022AF79CBF blob_commitment=1cb9db54a2d51ef2f219bf3138212f0b981f31f2f61f6dcf3137d4bdaf8859f9 rows_count=16384 error="rpc error: code = InvalidArgument desc = payment promise verification failed: stateful validation request: rpc error: code = Internal desc = creation_timestamp 2025-11-05 19:12:27.556169 +0000 UTC is greater than current timestamp 2025-11-05 19:12:27.236845 +0000 UTC" 
```

For now, I'd recommend setting the timestamps to time.Now() - 10 seconds

Longer term I would recommend [this](https://github.com/celestiaorg/celestia-app-fibre/issues/60#issuecomment-3492316104) path

### LINKED ISSUE #107: PaymentPromise creation timestamp always ahead of block time
## Context

Spun out from https://github.com/celestiaorg/celestia-app-fibre/pull/106

## Problem

The PaymentPromise creation time was always ahead of the current block time which results in an error

```
2025/11/21 10:22:08 WARN failed to upload rows fibre-client.validator=B0905579099988703E7C655E6695E161317D8479 fibre-client.blob_commitment=2b07ad6c3d685a75d10750385c2691592a5e456c6d9a9d0e8eb29a2d9df6bf9a fibre-client.rows_count=16384 fibre-client.error="rpc error: code = InvalidArgument desc = payment promise verification failed: stateful validation request: rpc error: code = Internal desc = creation_timestamp 2025-11-21 15:22:08.549773 +0000 UTC is greater than current timestamp 2025-11-21 15:22:07.195945 +0000 UTC"
```

The bug is here

https://github.com/celestiaorg/celestia-app-fibre/blob/6efa58d27c617b7a39ea2769a2a6af965afb3ca2/fibre/client_upload.go#L155

## Proposal

Modify the client to get the latest block time and use that as the PaymentPromise creation time.

rootulp:
Closes https://github.com/celestiaorg/celestia-app-fibre/issues/107
Closes https://github.com/celestiaorg/celestia-app-fibre/issues/67

### DIFF
--- a/fibre/client.go
+++ b/fibre/client.go
@@ -1,19 +1,16 @@
 package fibre
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"log/slog"
 	"sync"
 	"sync/atomic"
-	"time"
 
 	"github.com/celestiaorg/celestia-app/v6/fibre/grpc"
 	"github.com/celestiaorg/celestia-app/v6/fibre/validator"
 	"github.com/celestiaorg/celestia-app/v6/pkg/user"
 	cmtmath "github.com/cometbft/cometbft/libs/math"
-	tmservice "github.com/cosmos/cosmos-sdk/client/grpc/cmtservice"
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
 	clock "github.com/filecoin-project/go-clock"
 	"go.opentelemetry.io/otel"
@@ -161,24 +158,4 @@ func (c *Client) Close() error {
 func MaxMessageSize(cfg BlobConfig) int {
 	msgSize := cfg.MaxShardSize() + MaxPaymentPromiseSize
 	return msgSize + (msgSize / 50) // add 2% protobuf overhead
-}
-
-// latestBlockTime fetches the latest block time from the chain.
-func (c *Client) latestBlockTime(ctx context.Context) (time.Time, error) {
-	conn := c.txClient.PrimaryConn()
-	if conn == nil {
-		return time.Time{}, fmt.Errorf("no grpc connection available")
-	}
-
-	client := tmservice.NewServiceClient(conn)
-	resp, err := client.GetLatestBlock(ctx, &tmservice.GetLatestBlockRequest{})
-	if err != nil {
-		return time.Time{}, fmt.Errorf("failed to get latest block: %w", err)
-	}
-
-	if resp == nil || resp.SdkBlock == nil {
-		return time.Time{}, fmt.Errorf("invalid block response: missing block")
-	}
-
-	return resp.SdkBlock.Header.Time, nil
 }
--- a/fibre/client.go
+++ b/fibre/client.go
@@ -1,16 +1,19 @@
 package fibre
 
 import (
+	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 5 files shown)


Wondertan (dismissed):
This adds a roundtrip for no good reason. We should instead revisit the need for the check in the module this fix tried to mitigate. The check enforces that PP timestamp is less then block's, which is an unrealistic rule. Furthermore, server similarly checks PP timestamp, so the check in the module can be removed without any repercussions.

rootulp:
Can we merge this fix and revisit the design in parallel?

How did ya'll do benchmarking without a fix like this?

Wondertan:
There is nothing to be designed here. We dont need to add more code that adds latency and complexity when there is a single if statement that can be just removed and have the same exact effect.

evan-forbes:
> How did ya'll do benchmarking without a fix like this?

we temporarily disabled the time check for PFFs, as we were discussing to use heights etc. 

and then ultimately just didn't verfiy stateful things at all

with the new development of v7 very possibly not having consistent block times, we should likely avoid relying on height as a proxy for time. 

we can't enable folks use a time in the future for their promises, so agree that we should just use the time from the header. There are a few different ways to optimize that in the future. We could include this when we query for the validator set, or we could even pass this in from a light client.

Since removing round trips is an optimization, I think it okay to flup and not worth blocking. we can even label the flup with a `latency` tag to easily find optimizations for latency. It will also be easier to optimize that after the system is more complete.

evan-forbes:
@Wondertan 

> The check enforces that PP timestamp is less then block's, which is an unrealistic rul

if I'm understanding correctly, you are saying that we don't need to check the time at all, is that correct? if so, how do we prevent someone from setting a time very far in the future and avoiding any pruning?

Wondertan:
> Furthermore, server similarly checks PP timestamp, so the check in the module can be removed without any repercussions.


@evan-forbes

Wondertan:
> we can't enable folks use a time in the future for their promises, so agree that we should just use the time from the header. There are a few different ways to optimize that in the future. We could include this when we query for the validator set, or we could even pass this in from a light client.


Server uses time.Now allowing for a drift which is exactly how time in header verification is checked too.

evan-forbes:
okay that make sense

we can remove the check then. subtracting a second or something should guarantee the server accepts the value

rootulp:
So the concrete follow-up here is to remove the conditional on the CreationTimestamp inside the x/fibre module.

rootulp:
Unclear to me if x/fibre should be validating the payment promise's creation timestamp is not too old. But the script doesn't hit an issue there so we can FLUP if that conditional also needs to be removed.

evan-forbes (approved):
dope thanks @rootulp and @Wondertan 

I think we got a good outcome here

PR #110, opened 2025-11-21
fix(fibre): provide a unique ID to `RawBytesMessageSignBytes`

### LINKED ISSUE #109: `error signing raw bytes: uniqueID cannot be empty`
## Context

Spun out from https://github.com/celestiaorg/celestia-app-fibre/pull/106

## Problem

The fibre code invokes `SignRawBytes` with a unique ID of `""` which fails validation and results in this error

```
2025/11/21 11:27:52 WARN failed to upload rows fibre-client.validator=5D338810A780DD77CCC936485F118AC957ACB7FD fibre-client.blob_commitment=703e5bad7f2fc5a11dd58e167f06496f82549ee17a544826c7c71ebc4b9f488d fibre-client.rows_count=16384 fibre-client.error="rpc error: code = Internal desc = failed to sign payment promise: signing payment promise: error signing raw bytes: uniqueID cannot be empty"
```

The bug is here
https://github.com/celestiaorg/celestia-app-fibre/blob/0c4c1e0170fdf96a29c226e9826408f14dbd445f/fibre/server_upload.go#L237

## Problem

Update that code to pass a unique ID. Based on celestia-core it is possible to use a string constant like `"compactBlock"` as the unique ID so proposal to use `"paymentPromise"`.

After making that change, we'll need to update the signature verification logic accordingly.

### LINKED ISSUE #111: Don’t pass an empty string to the Privval when signing valid promises
Don’t pass an empty string to the Privval when signing valid promises
Currently we are passing a uniqueID of “” when signing with the Privval, the Privval currently doesn’t allow for this.
https://github.com/celestiaorg/celestia-app-fibre/commit/cce8cff5d7ef6d1b88adef5e0e992934bf84138f
https://github.com/celestiaorg/celestia-app-fibre/commit/c8d5e31af82252cbf76c59d032721b6bfe7b83a7

rootulp:
Closes https://github.com/celestiaorg/celestia-app-fibre/issues/109
Closes https://github.com/celestiaorg/celestia-app-fibre/issues/111

### DIFF
--- a/fibre/server_upload.go
+++ b/fibre/server_upload.go
@@ -234,8 +234,7 @@ func (s *Server) signPromise(promise *PaymentPromise) ([]byte, error) {
 	}
 
 	// sign using validator's private key
-	// SignRawBytes constructs bytes as: chainID || uniqueID || rawBytes
-	signature, err := s.privVal.SignRawBytes(s.cfg.ChainID, ValidatorSignatureUniqueID, signBytes)
+	signature, err := s.privVal.SignRawBytes(s.cfg.ChainID, "", signBytes) // signBytes already include domain separator, so we don't have to pass its
 	if err != nil {
 		return nil, fmt.Errorf("signing payment promise: %w", err)
 	}
--- a/fibre/server_upload.go
+++ b/fibre/server_upload.go
@@ -18,11 +18,6 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-const (
-	// PaymentPromiseUID is the unique ID used for payment promise signing.
-	PaymentPromiseUID = "paymentPromise"
-)
-
 // UploadShard handles the [types.FibreServer.UploadShard] RPC call.
 func (s *Server) UploadShard(ctx context.Context, req *types.UploadShardRequest) (*types.UploadShardResponse, error) {
 	ctx, span := s.tracer.Start(ctx, "fibre.Server.UploadShard")
@@ -239,7 +234,7 @@ func (s *Server) signPromise(promise *PaymentPromise) ([]byte, error) {
 	}
 
 	// sign using validator's private key
-	signature, err := s.privVal.SignRawBytes(s.cfg.ChainID, PaymentPromiseUID, signBytes) 
+	signature, err := s.privVal.SignRawBytes(s.cfg.ChainID, SignBytesPrefix, signBytes)
 	if err != nil {
 		return nil, fmt.Errorf("signing payment promise: %w", err)
 	}
--- a/fibre/server_upload.go
+++ b/fibre/server_upload.go
@@ -239,7 +239,7 @@ func (s *Server) signPromise(promise *PaymentPromise) ([]byte, error) {
 	}
 
 	// sign using validator's private key
-	signature, err := s.privVal.SignRawBytes(s.cfg.ChainID, PaymentPromiseUID, signBytes) // signBytes already include domain separator, so we don't have to pass its
+	signature, err := s.privVal.SignRawBytes(s.cfg.ChainID, PaymentPromiseUID, signBytes) 
 	if err != nil {
 		return nil, fmt.Errorf("signing payment promise: %w", err)
 	}
--- a/fibre/server_upload.go
+++ b/fibre/server_upload.go
@@ -234,7 +234,8 @@ func (s *Server) signPromise(promise *PaymentPromise) ([]byte, error) {
 	}
 
 	// sign using validator's private key
-	signature, err := s.privVal.SignRawBytes(s.cfg.ChainID, "", signBytes) // signBytes already include domain separator, so we don't have to pass its
+	// SignRawBytes constructs bytes as: chainID || uniqueID || rawBytes
+	signature, err := s.privVal.SignRawBytes(s.cfg.ChainID, ValidatorSignatureUniqueID, signBytes)
 	if err != nil {
 		return nil, fmt.Errorf("signing payment promise: %w", err)
 	}
--- a/fibre/server_upload.go
+++ b/fibre/server_upload.go
@@ -18,6 +18,11 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+const (
+	// PaymentPromiseUID is the unique ID used for payment promise signing.
+	PaymentPromiseUID = "paymentPromise"
+)
+
 // UploadShard handles the [types.FibreServer.UploadShard] RPC call.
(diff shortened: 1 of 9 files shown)


Wondertan (commented):
Is there a way to fix the signer interface instead? Both chain id and domain separator are included in the sign bytes already. Now, we will have to also specify this rule additionally on top. Essentially, this change adds techdebt for both protocol and implementation levels, as will have to clean this up at some point and its also breaking, while there is nothing wrong with signing over pure bytes and enforcing domain separation all the time is misguided.

fibre/server_upload.go
```diff
@@ -18,6 +18,11 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+const (
+	// PaymentPromiseUID is the unique ID used for payment promise signing.
+	PaymentPromiseUID = "paymentPromise"
```
evan-forbes: while maxxing we used a slightly different unique id here

the only thing I'd recommend is to add a version, such as paymentPromise-v1

evan-forbes (commented):
also closes #111 apologies didn't see you also opened this issue while I was writing up all the stuff we found while maxxxing

evan-forbes:
@Wondertan the extra domain separator here is actually from upstream comet, they wanted it to be extra explicit although I agree it is not needed

any changes we make to the Privval also means updating all of the implementations there such as tmkms and horcrux, both of which follow the cometBFT interface and reject empty strings afaiu

rootulp:
> @Wondertan the extra domain separator here is actually from upstream comet

+1. We can't use `privVal.SignRawBytes` with a "" unique ID. It seems pretty clear that this code never worked as intended and this PR fixes the bug.

https://github.com/celestiaorg/celestia-app-fibre/blob/0c4c1e0170fdf96a29c226e9826408f14dbd445f/fibre/server_upload.go#L237

fibre/server_upload.go
rootulp: can do

rootulp:
FWIW it's extremely demotivating to get blocked on something like this: https://github.com/celestiaorg/celestia-app-fibre/pull/110#pullrequestreview-3494293029

Also if you used a similar fix for benchmarking, why is that fix no longer acceptable?

fibre/server_upload.go
```diff
@@ -234,7 +239,7 @@ func (s *Server) signPromise(promise *PaymentPromise) ([]byte, error) {
 	}
 
 	// sign using validator's private key
-	signature, err := s.privVal.SignRawBytes(s.cfg.ChainID, "", signBytes) // signBytes already include domain separator, so we don't have to pass its
+	signature, err := s.privVal.SignRawBytes(s.cfg.ChainID, PaymentPromiseUID, signBytes) // signBytes already include domain separator, so we don't have to pass its
```
Wondertan: The comment needs to be updated, as apparently domain separator is enforced.

Wondertan:
Not exactly sure whats exactly demotivating about my question and sharing my understanding of the change consequences, but its not blocking as I didn't requested changes.

fibre/server_upload.go
Wondertan: We already have it defined in constants next to payment promise and also with versioning. There is no need for another constant.

fibre/validator/signature_set.go
```diff
@@ -38,9 +42,15 @@ func (s Set) NewSignatureSet(targetVotingPower, targetValidatorsCount cmtmath.Fr
 // Returns an error if the signature is invalid.
 // Returns true if enough signatures have been collected to meet both thresholds.
 func (ss *SignatureSet) Add(val *core.Validator, signature []byte) (bool, error) {
+	// reconstruct the signed message using RawBytesMessageSignBytes
+	signBytes, err := core.RawBytesMessageSignBytes(ss.chainID, ss.uniqueID, ss.requiredBytesSigned)
+	if err != nil {
+		return false, fmt.Errorf("failed to reconstruct sign bytes: %w", err)
+	}
```
Wondertan: Please prepare sign bytes outside of SignatureSet and simply pass the expected bytes just like before. The caller there has access to constant already defined for domain separation as well as has chain id handy.

Wondertan (dismissed):
However, now its blocking. Apparently fixing the comets privval interface is too big of a hussle and we need to stick with techdebt and specing out this nuance, so here is actual review for this change. The main concern is defining an additionally domain separator constant.

fibre/validator/signature_set.go
Wondertan: ~~And even better option that should work for the module too is to put this inside pp.SignBytes method. PP has all the fields on it already and then we dont need to touch SingatureSet at all to fix this + we keep the sign bytes exactly as they are expected to be for the verifier.~~

Ah well no, it wont work, because the call to Privval does this operation internally on top, so it would be done twice while the verifier expects it only once. Annoying that this interface is not symmetrical...

Sticking to op then

fibre/server_upload.go
rootulp:
```suggestion
	signature, err := s.privVal.SignRawBytes(s.cfg.ChainID, PaymentPromiseUID, signBytes) 
```

fibre/server_upload.go
rootulp: Where? A link would be more helpful.

rootulp: Found this and used it: 11782ce1717378aab31ffd0bec0617726da6fa5b

fibre/validator/signature_set_test.go
```diff
@@ -1,6 +1,7 @@
 package validator_test
```
Wondertan, line 1: Changes in this file are unnecessary; it should sign well as before without additional sign bytes preparation. The SignatureSet can work withwhatever signbytes given to it and enforcement of particular format is a concern of a layer above, not of the SignatureSet

fibre/validator/signature_set.go
```diff
@@ -22,6 +22,8 @@ type SignatureSet struct {
 }
 
 // NewSignatureSet creates a new [SignatureSet] for collecting and validating signatures.
+// The requiredBytesSigned parameter should contain the already-prepared sign bytes
+// (with domain separation prefix and chainID) that validators sign over.
```
Wondertan: nit: this is true from privval perspective; however, locally within the signature set, this is not the case, and it can work with whatever signbytes given to it, so I would just remove this comment change.

Wondertan (commented):
almost there

rootulp:
make test-short is failing but the failures don't appear unique to this PR: https://github.com/celestiaorg/celestia-app-fibre/issues/121

evan-forbes (approved):

PR #133, opened 2025-12-08
refactor(fibre): consolidate PP validator signing

### LINKED ISSUE #128: While we on it, i would iterate on the idea of making  this a method on PP. Not a part of existing SignBytes (which wouldnt work as explained in the previous PR) but a new additional one. (E.g. ValidatorSignBytes)
While we on it, i would iterate on the idea of making  this a method on PP. Not a part of existing SignBytes (which wouldnt work as explained in the previous PR) but a new additional one. (E.g. ValidatorSignBytes)

This wont just be a change to improve ergonomics for accessing bytes to sign, but will also keep this signing quirk contained in a single spot, preventing misuse and making it easy to evolve/change the PP. The users wouldnt need to import the core pkg, pass anything to this hacky function, i.e. the complexity of understanding what to sign is abstracted. Oh, and also the SignBytesPrefix can then be back privatised.

_Originally posted by @Wondertan in https://github.com/celestiaorg/celestia-app-fibre/pull/126#discussion_r2561463628_

Wondertan:
No functional changes expected. Additionally, hardness and seals the signing bug story with a unit test.

Closes #128


fibre/payment_promise.go
```diff
@@ -267,3 +269,35 @@ func (p *PaymentPromise) Hash() ([]byte, error) {
 	}
 	return p.hash[:], nil
 }
+
+// SignBytesValidator returns the [PaymentPromise] bytes for validators to sign.
+// This wraps the [PaymentPromise.SignBytes] with domain separation using the chain ID and signBytesPrefix.
+//
+// NOTE: This method encapsulates Comet's quirk which enforces a particular signing domain separation format.
```
cmwaters, line 276: [We introduced it I believe]

fibre/payment_promise.go
```diff
@@ -267,3 +269,35 @@ func (p *PaymentPromise) Hash() ([]byte, error) {
...
+// This goes on top of native [PaymentPromise] domain separation.
+func (p *PaymentPromise) SignBytesValidator() ([]byte, error) {
+	signBytes, err := p.SignBytes()
```
cmwaters, line 279: nit: Maybe `SignBytes` can be renamed `RawSignBytes` or something to avoid the confusion that these are the bytes that actually get signed over. Or maybe the method should be internal.

cmwaters (dismissed):
LGTM

fibre/payment_promise.go
Wondertan: SignBytes is used to sign by the submitter, so it shouldn't be internal.

fibre/payment_promise.go
Wondertan: PrivVal interface enforces this. If we introduced this to the interface then i am blaming us and not comet team for this "design decision".

walldiss (dismissed):

fibre/client_upload.go
```diff
@@ -70,15 +70,7 @@ func (c *Client) Upload(ctx context.Context, ns share.Namespace, blob *Blob) (re
...
-	signBytes, err := promise.SignBytes()
-	if err != nil {
-		span.RecordError(err)
-		span.SetStatus(codes.Error, "failed to prepare bytes to sign")
-		return result, fmt.Errorf("preparing bytes to sign: %w", err)
-	}
-
-	// Prepare sign bytes with domain separation and chainID for validator signatures
-	validatorSignBytes, err := core.RawBytesMessageSignBytes(c.cfg.ChainID, SignBytesPrefix, signBytes)
+	validatorSignBytes, err := promise.SignBytesValidator()
```
rootulp, line 73: [nit] Can we enforce a consistent order for this name? Either:

```go
	validatorSignBytes, err := promise.ValidatorSignBytes()
```

or

```go
	signBytesValidator, err := promise.SignBytesValidator()
```

rootulp (approved):

PR #134, opened 2025-12-08
fix(fibre/validator): overassigning overlaps

Wondertan:
There were possible row overlaps when assigning. By following the diff you will see why.


cmwaters (approved):
Nice find. I am assuming `TestSet_Assign` was failing before

Wondertan:
It wasn't failing. With the test changes in this PR it would fail without respective fix in the code.

Wondertan:
It wasn't failing. With the test changes in this PR it would fail without respective fix in the code.

walldiss (approved):
Nice find!

PR #138, opened 2025-12-11
feat(fibre/server): time-based pruning

Wondertan:
Adds `PruneBefore` method into Store and the pruning loop in Server that triggers the method every minute.

The pruning logic was benchmarked and optimized to the point that it iterates only over the entries to remove, leveraging the lexicographic ordering of keys in the index. 

The pruning index is updated to keep timestamps, not the PP creation time, but the time the data is supposed to be removed, as reported by the state module (future change).

### DIFF
--- a/cmd/celestia-appd/cmd/start_command_standalone.go
+++ b/cmd/celestia-appd/cmd/start_command_standalone.go
@@ -109,6 +109,8 @@ func startCommandHandler(
 		if err != nil {
 			return fmt.Errorf("failed to start Fibre server: %w", err)
 		}
+		fibreServer.Start()
+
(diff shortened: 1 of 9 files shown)


walldiss (dismissed):
Looks good! 
Clarified with @Wondertan that changes to time based buckets were necessary to keep pruning window constant definition/governance inside state machine. 

Would be interesting to benchmarks over filled storage to make sure performance is sufficient when indexes are full

Wondertan:
> Would be interesting to benchmarks over filled storage to make sure performance is sufficient when indexes are full

I believe current benchmark covers this case. You can control the percentage of indexies to be pruned out of total entries and its scales well. O(n) where n is number of entries to be pruned.

walldiss (approved):

PR #141, opened 2025-12-17
refactor: use ValidatePaymentPromise for stateful verification

### LINKED ISSUE #60: Remove block time from Fibre server
## Context

Discussion with Hlib.

## Problem

The Fibre server currently needs a `BlockTime`. Since celestia-app can't guarantee a block time I infer that all operations that the Fibre server performs using this block time can't be 100% accurate because the actual block time may deviate from the block time configured on the Fibre server (e.g. 6 seconds).

## Proposal

1. Investigate what operations the Fibre server uses the `BlockTime` for. If none of the operations are critical, close this issue as won't do.
2. If any of the operations become invalid when the observed block time differs from the server configured block time then we need to mitigate. Ideally the Fibre server doesn't need a `BlockTime` config.

### LINKED ISSUE #129: Use `ValidatePaymentPromise` for all stateful verification of promises
### Summary

There are a few areas where we are using constants in the fibre server to verify aspects of the promise where we should be using the stateful `ValidatePaymentPromise` method. 

https://github.com/celestiaorg/celestia-app-fibre/blob/6f9be0c390b018182415f4f5386dc583d7b1e917/fibre/server_upload.go#L111-L128

`PaymentPromiseTimeout` and `BlockTime` should be removed from the config as these are not constants but a variables and can change. It will be more robust to instead use `ValidatePaymentPromise` to verify the height and the timestamp. 

The `ValidatePaymentPromise` method should return the expiration time

rootulp:
Closes https://github.com/celestiaorg/celestia-app-fibre/issues/60
Closes https://github.com/celestiaorg/celestia-app-fibre/issues/129

## Motivation

- https://www.notion.so/celestiaorg/Hlib-Cal-Timestamp-vs-Height-2bd2a70c951580e3a6dff1b5d063480b
- https://www.notion.so/celestiaorg/08-12-25-2c32a70c951580d1b9c4f2f46979304b?source=copy_link#2c32a70c9515803d975afaed9b993c54

## Description

- Adds a new governance modifiable param for the block height range (defaults to 1000). Payment promises must have a height in this range: `[currentHeight-1000, currentHeight+1]`


ninabarbakadze (commented):
merge conflict

fibre/server_upload.go
```diff
@@ -94,41 +95,44 @@ func (s *Server) UploadShard(ctx context.Context, req *types.UploadShardRequest)
...
 	// validate PP fields matches the config
 	if promise.ChainID != s.cfg.ChainID {
-		return nil, nil, fmt.Errorf("payment promise chain ID mismatch: expected %s, got %s", s.cfg.ChainID, promise.ChainID)
+		return nil, nil, time.Time{}, fmt.Errorf("payment promise chain ID mismatch: expected %s, got %s", s.cfg.ChainID, promise.ChainID)
 	}
 	if promise.BlobVersion != uint32(s.cfg.BlobVersion) {
-		return nil, nil, fmt.Errorf("blob version mismatch: expected %d, got %d", s.cfg.BlobVersion, promise.BlobVersion)
+		return nil, nil, time.Time{}, fmt.Errorf("blob version mismatch: expected %d, got %d", s.cfg.BlobVersion, promise.BlobVersion)
 	}
+	pruneAt := time.Now().Add(s.cfg.DataRetentionDuration)
```
Wondertan: Should be the expiration time!

Wondertan: Also, can remove `DataRetentionDuration` as it's now governed in the chain state.

fibre/server_upload.go
```diff
@@ -109,27 +109,6 @@ func (s *Server) verifyPromise(ctx context.Context, promisePb *types.PaymentProm
...
-				s.cfg.PaymentPromiseTimeout)
-		}
-		// use height of the latest valset to verify height in the promise
-		currentValSet, err := s.valGet.Head(ctx)
-		if err != nil {
-			return nil, nil, time.Time{}, fmt.Errorf("getting current validator set: %w", err)
-		}
-		// calculate max height drift based on promise timeout and block time
-		maxHeightDrift := uint64(s.cfg.PaymentPromiseTimeout / s.cfg.BlockTime)
-		if currentValSet.Height > maxHeightDrift && promise.Height < currentValSet.Height-maxHeightDrift {
```
Wondertan, line 128: We still need to check the height. Couldn't find one added to the state.

Wondertan (changes requested):

fibre/server_upload.go
Wondertan: and `BlockTime`

fibre/server_upload.go
rootulp: Check the height against what? Is the height check specified anywhere? I don't see it in https://github.com/celestiaorg/fibre-da-spec/blob/main/server.md#2-handler-semantics

fibre/server_upload.go
rootulp: a3e412074d9495178428c75ca5d9d58013949a60

fibre/server_upload.go
rootulp: Given your other comment suggested removing BlockTime from the config, I don't see how we can keep the current height check.

Wondertan: I mean there should be some form of validity check for the height. It doesn't have to be block time based and likely another governance controlled parameter.

Wondertan:
Сloses #60 now too

fibre/server_upload.go
rootulp: https://github.com/celestiaorg/celestia-app-fibre/pull/141/commits/cbeca470ffb7945c1a08e1cfbe67893beeca4016

x/fibre/keeper/keeper.go
```diff
@@ -310,40 +307,55 @@ func (k Keeper) ParseProcessedPaymentsByTimeKey(key []byte) (processedAt time.Ti
...
-// This method does NOT perform stateless validation.
-// Callers should perform stateless validation separately via pp.Validate().
-func (k Keeper) ValidatePaymentPromiseStateful(ctx sdk.Context, promise *types.PaymentPromise) error {
-	// Validate creation_timestamp bounds
-	// Spec requirement: creation_timestamp <= current confirmed timestamp
-	// and creation_timestamp > (header_timestamp - withdrawal_delay)
+// validatePaymentPromiseStatefulInternal performs the core stateful validation logic.
+// The isTimeout parameter indicates whether this is being called for timeout processing,
+// which skips expiration and height validation to allow processing older promises.
+func (k Keeper) validatePaymentPromiseStatefulInternal(ctx sdk.Context, promise *types.PaymentPromise, isTimeout bool) (time.Time, error) {
```
Wondertan, line 313: personally not a fan of passing a flag this way when this could be refactored into two distinct methods where one calls(wraps/embeds) the other with more conditions, but not a big deal to keep as is tbf.

Wondertan (approved):
The server and keeper logic LGTM. Thank you

PR #145, opened 2026-01-09
refactor(fibre): extract protocol params

Wondertan:
The premise behind this change started from work on https://github.com/celestiaorg/celestia-app-fibre/issues/80. The current parameter configuration was overloaded and convoluted, and more dynamic and static configurations were about to be added for #80, which together called for a cleanup. 

In the end, in this PR, existing protocol parameters are extracted, and new ones are added into a separate new ProtocolParams struct that defines fundamental protocol constants from which other parameters are calculated and used in the respective blob, client, and server configs.

Besides, a MaxBlobSize check bug was fixed, in which the blob header overhead was not taken into account, and the first steps towards blob versioning were taken.

The newly added and existing parameters were audited with @nashqueue.

PR #146, opened 2026-01-09
perf(fibre/validator): assign less rows

### LINKED ISSUE #80: fibre/validator: Send only necessary amount of row
Currently, the `Assign` function assigns all the rows to all validators uniformly. Instead, we should calculate the minimum number of rows per validator needed, given a 1:3 encoding. 

In the `BlobConfig`, we should add one more method that calculates `RowsPerShard() int` based on existing parameters and then we pass it as a new additional param to `Assign`

Wondertan:
The actual formula for how many rows to send is in #145, and this PR only changes the usage

Depends on #145
Closes https://github.com/celestiaorg/celestia-app-fibre/issues/80


rootulp (dismissed):
CI is failing

rootulp:
CI is still failing. You can ignore modernize-check failing. I fixed it on celestia-app repo by upgrading to Go 1.25.5 and adding modernize to golangci-lint instead of a separate script. 

Ref: https://github.com/celestiaorg/celestia-app/pull/6361

but test and lint are failing.

Wondertan:
Yeah, I know, I only rebased.

Wondertan:
The problem with the linter is unrelated to this PR btw

Wondertan:
and [failed the same way ](https://github.com/celestiaorg/celestia-app-fibre/actions/runs/20917970026/job/60096135693)for other PRs you approved

Wondertan:
Also, the test short fails because a 1m timeout is enforced, and the maxblob test needs more time than that on the CI's hardware. Not sure why wouldn't run the full test suite tho. I know it takes time, but we its worth catching integration tests failures on CI too

rootulp:
> Also, the test short fails because a 1m timeout is enforced, and the maxblob test needs more time than that on the CI's hardware. Not sure why wouldn't run the full test suite tho. I know it takes time, but we its worth catching integration tests failures on CI too

test-short commands are limited to 1m so that developers have a fast iteration cycle locally. If a test takes longer than 1m it should be ignored from running in short mode. test still runs in CI so even if the test is ignored in short mode, it will still run in test.

Wondertan:
This PR is ready for review, @rootulp. Both linter and test issues are out of scope for this PR and existed before it.

Wondertan:
> test-short commands are limited to 1m so that developers have a fast iteration cycle locally.

Yeah, locally, which means there is no point in running a CI job for it, since the overall test runner includes all the short tests and additional short test will only make CI longer for no reason.

rootulp:
>Yeah, locally, which means there is no point in running a CI job for it, since the overall test runner includes all the short tests and additional short test will only make CI longer for no reason.

The point of running it in CI is to catch cases where a developer adds a long running test and fails to skip it in short mode. In other words, test-short in CI is serving it's purpose right now. It is blocking this PR from merging because a long running test isn't skipped in short mode:

```go
func TestClientServerUpload(t *testing.T) {
	if testing.Short() {
		t.Skip("skipping TestClientServerUpload in short mode")
	}
```

Update: 
- https://github.com/celestiaorg/celestia-app-fibre/pull/153
- https://github.com/celestiaorg/celestia-app-fibre/pull/154
- https://github.com/celestiaorg/celestia-app-fibre/pull/155

Wondertan:
Ok, the intention behind test-short on CI is now clear to me. One thing to note is that a test that runs long in CI might not run as long as locally due to hardware differences.

Wondertan:
@rootulp, if you want to fix all the existing CI issues before emerging this PR, there is one more. The test race fails, although all the tests succeed there as seen in action logs

rootulp (approved):

Issue #144, opened 2026-01-09
distribute the rows by stake instead equally


liamsi:
see: https://github.com/celestiaorg/celestia-app/issues/6666

PR #148, opened 2026-01-12
refactor: remove 2/3 signature count

### LINKED ISSUE #147: Update validator signature verification to remove 2/3 count
## Context

https://github.com/celestiaorg/celestia-app-fibre/issues/144

## Problem

https://github.com/celestiaorg/celestia-app-fibre/blob/4c83a2153a880f83b1867098849ec1f314103194/fibre/validator/signature_set.go#L65-L74

countNotMet is no longer required

## Proposal

Remove countNotMet. Audit if there are any other instances where validator signature count is checked because it is no longer needed

rootulp:
Closes https://github.com/celestiaorg/celestia-app-fibre/issues/147

We may want to hold off on merging this until other PRs are ready for https://github.com/celestiaorg/celestia-app-fibre/issues/144

### DIFF
--- a/fibre/client.go
+++ b/fibre/client.go
@@ -40,8 +40,6 @@ type ClientConfig struct {
 
 	// UploadTargetVotingPower is the fraction (e.g., 2/3) of total voting power required for Upload operations.
 	UploadTargetVotingPower cmtmath.Fraction
-	// UploadTargetSignaturesCount is the fraction (e.g., 2/3) of total signature count required for Upload operations.
-	UploadTargetSignaturesCount cmtmath.Fraction
 	// UploadConcurrency is the maximum number of concurrent uploads to validators.
 	UploadConcurrency int
 	// DownloadConcurrency is the maximum number of concurrent read requests to validators.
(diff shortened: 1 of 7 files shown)


Wondertan:
Afaik, we will need this functionality for pog thats confirmed to happen, so I would also consider adding a simple flag to disable signatures count instead of removing it entirely

rootulp:
I don't understand why PoG requires a 2/3 signature count check, can you please elaborate or point to resources that explain why?

Wondertan:
Pog implies equal votes amongst validators meaning stake becomes irrelevant and only 2/3 signatures like we remove here matters

rootulp:
In PoG if all all validators have equal stake then we only need to check 2/3 of stake which is equivalent to checking 2/3 of count. It seems valid to check 2/3 stake or 2/3 count. Without more details on PoG implementation we don't know if we need a conditional here. Seems easy to add later based on git history anyway so inclined to merge this as-is.

Wondertan (approved):

rootulp:
I didn't close this PR, not sure why Github closed it on my behalf. I can't re-open this one.

PR #151, opened 2026-01-13
chore(fibre/store): add synthetic benchmarks

Wondertan:
Adds synthetic benchmarks for the store to establish throughput baselines and compare database backends (Badger vs Pebble). Note that this is a theoretical best-case baseline that doesn't measure throughput during compaction, or any sustained load with read/write contentions. 

Both backends performed similarly, achieving ~200 MiB/s of goodput at the maximum blob size. Reads were ~3-4 times faster than writes for both, but Pebble was slightly faster overall. Badger was slightly better for large blobs (+8%), and Pebble was slightly better for medium blobs (+19% write at 16 MiB). Concurrent writes improve throughput by ~37% (peak at 4-8 workers) for both.

Given how similar they performed, it's hard to decide on the backend yet. I think we still need to run a sustained e2e load test first, as this benchmark isn't sufficiently representative of the production environment. It also might be that my laptop's disk is a bottleneck, so they performed similarly.

Assumes equal stake distribution. 

Depends on #145


cmwaters (dismissed):
Thanks for doing these preliminary benchmarks. 

I think we should merge this as is and then discuss what optimizations we want to make prior to release and what backend we want to choose. I think for simplicity and uniformity we should converge on a single backend

Wondertan:
Need to fix the lint

evan-forbes (approved):
👍

PR #155, opened 2026-01-16
chore: remove modernize check

rootulp:
Its failing on this repo because of Go versions. I fixed on celestia-app by updating to Go 1.25.5 but that doesn't seem strictly necessary here. Opted to just remove it from this fork for now.


Wondertan (approved):
Updating go version seems to be an easier thing to do, but up to you

rootulp:
it was a larger diff so opted for removing from now.

PR #156, opened 2026-01-16
feat(fibre/validator): stake-based assign

### LINKED ISSUE #144: distribute the rows by stake instead equally

Wondertan:
Notably had to also change the MaxMessageSize calculation and bring in another assumption that no validator can have more than 1-SafetyThreshold power as stake power ceiling for row distribution for a validator. 

Depends on #146 
Closes #144

### DIFF
--- a/fibre/client.go
+++ b/fibre/client.go
@@ -35,13 +35,16 @@ type ClientConfig struct {
 	// ChainID is the chain identifier for domain separation in [PaymentPromise] signatures.
 	ChainID string
 
-	// RowsPerShard computes the number of rows per shard given the total number of shards.
-	RowsPerShard func(totalShards int) int
+	// SafetyThreshold is the fraction of stake needed to cause a safety failure (typically 2/3).
+	SafetyThreshold cmtmath.Fraction
+	// LivenessThreshold is the fraction of stake needed to cause a liveness failure (typically 1/3).
+	LivenessThreshold cmtmath.Fraction
+	// MinRowsPerValidator is the minimum number of rows each validator must receive
+	// for unique decodability security.
+	MinRowsPerValidator int
 	// MaxMessageSize is the maximum gRPC message size for upload requests.
 	MaxMessageSize int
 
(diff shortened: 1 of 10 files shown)


fibre/client.go
```diff
@@ -35,8 +35,9 @@ type ClientConfig struct {
 	// ChainID is the chain identifier for domain separation in [PaymentPromise] signatures.
 	ChainID string
 
-	// RowsPerShard computes the number of rows per shard given the total number of shards.
-	RowsPerShard func(totalShards int) int
+	// MinRowsPerShard computes the minimum number of rows each validator must receive
+	// given the total number of shards, for unique decodability security.
+	MinRowsPerShard func(totalShards int) int
```
rootulp: [nit] I still think it's confusing that we use "shard" here. The comment already uses the term validator and I think the original intention was to not use validator because this wanted to be abstract of validators. Since this is inherently related to validators, we could remove a layer of indirection by renaming "shard" to "validator":

```suggestion
	// MinRowsPerValidator computes the minimum number of rows each validator must receive
	// given the total number of validator, for unique decodability security.
	MinRowsPerValidator func(totalValidators int) int
```

rootulp (dismissed):

fibre/client.go
Wondertan: True. Notions of shard and validator are interchangeable in this place but using validator does make it simpler to understand. I will update it.

Wondertan:
New push fixes [the comment](https://github.com/celestiaorg/celestia-app-fibre/pull/53#issuecomment-3772671180) from @nashqueue

nashqueue:
I did not know this math trick: 
rows := int((num + den - 1) / den) // ceil division

Assuming this is how you do ceil division, the assign function is mathematically correct.

Wondertan:
Its float free integer ceil devision

walldiss (approved):

walldiss (approved):

PR #161, opened 2026-01-21
feat(fibre/validator): add Select for stake-weighted validator selection

Wondertan:
## Summary

  Adds `validator.Set.Select` for stake-weighted validator selection during blob downloads.

  ### Design

  `Select` returns validators shuffled by stake weight for load balancing; higher-stake validators appear earlier, distributing download requests proportionally to capacity.

  **Duplicate handling**: When row assignments wrap around (due to `minRows` guarantees), some validators hold duplicate rows. These validators are placed at the end of the selection order and excluded from initial reconstruction requests. This is intentional: validators with duplicates are typically minimal-stake validators, which tend to be less reliable. They serve only as a fallback if initial requests fail.

  ### Returns

  - `validators`: Ordered slice for iterating through during download
  - `minRequired`: Number of validators needed for reconstruction (the split point between unique and duplicate holders)

  ### Testing

  - Added comprehensive coverage for `Select` behavior
  - Added tests verifying `Select` and `Assign` interact correctly
  - Tests cover edge cases: single validator, equal stakes, varying stake distributions

Extracted from #53 as it requires a more involved review. 
Depends on #156

### DIFF
--- a/fibre/validator/set.go
+++ b/fibre/validator/set.go
@@ -106,19 +106,18 @@ func (s Set) Select(originalRows, minRows int, livenessThreshold cmtmath.Fractio
 	validators = make([]*core.Validator, len(s.Validators))
 	copy(validators, s.Validators)
 
-	// minStakeFraction is the minimum contribution per validator for unique decodability
-	// e.g., 148 / (4096 * 3) ≈ 1.2% for livenessThreshold=1/3
-	totalDistributedRows := originalRows * int(livenessThreshold.Denominator) / int(livenessThreshold.Numerator)
-	minStakeFraction := float64(minRows) / float64(totalDistributedRows)
+	// find split point where row assignments start overlapping
+	// each validator contributes max(their stake, minStake) where minStake ensures unique decodability
+	totalStake := s.TotalVotingPower()
+	totalDistributedRows := int64(originalRows) * int64(livenessThreshold.Denominator) / int64(livenessThreshold.Numerator)
+	minStake := (int64(minRows)*totalStake + totalDistributedRows - 1) / totalDistributedRows // ceil division
 
-	// find last non-overlapping validator using effective stake (actual stake floored by minStakeFraction)
-	totalStake := float64(s.TotalVotingPower())
-	accumulated := 0.0
+	accumulated := int64(0)
 	splitIdx := len(validators)
 	for i, v := range validators {
-		accumulated += max(float64(v.VotingPower)/totalStake, minStakeFraction)
-		if accumulated >= 1.0 {
-			splitIdx = i + 1
+		accumulated += max(v.VotingPower, minStake)
+		if accumulated > totalStake {
+			splitIdx = i
 			break
 		}
 	}
--- a/fibre/validator/set.go
+++ b/fibre/validator/set.go
@@ -96,6 +96,78 @@ func (s Set) Assign(commitment rsema1d.Commitment, totalRows, originalRows, minR
 	return shardMap
 }
 
+// Select returns validators to download shards from, shuffled by stake for load balancing.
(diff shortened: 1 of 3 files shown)


rootulp (dismissed):

Wondertan:
Both #156  and #161 required a lot more effort in the end for coverage and all the edge cases, but should be good now.

fibre/validator/set.go
```diff
@@ -96,6 +96,78 @@ func (s Set) Assign(commitment rsema1d.Commitment, totalRows, originalRows, minR
...
+
+	// find last non-overlapping validator using effective stake (actual stake floored by minStakeFraction)
+	totalStake := float64(s.TotalVotingPower())
+	accumulated := 0.0
+	splitIdx := len(validators)
+	for i, v := range validators {
+		accumulated += max(float64(v.VotingPower)/totalStake, minStakeFraction)
+		if accumulated >= 1.0 {
+			splitIdx = i + 1
+			break
```
nashqueue: I think it would be best if you count >1 and not >=1, that way if we have exactly 1, there is no duplicate. 

Also, would return splitIdx = i
If you just discovered that with i validator you are over the total amount, then this validator is already part of the duplicates. Therefore, the validator should be on the right side of the slice.

fibre/validator/set.go
```diff
@@ -96,6 +96,78 @@ func (s Set) Assign(commitment rsema1d.Commitment, totalRows, originalRows, minR
...
+		}
+	}
+
+	// shuffle each group by stake for load balancing
+	// NOTE: doesn't require cryptographic randomness
+	rng := rand.New(rand.NewPCG(rand.Uint64(), rand.Uint64()))
+	shuffleByStake(validators[:splitIdx], rng)
+	shuffleByStake(validators[splitIdx:], rng)
```
nashqueue, line 129: As the shuffle is in-place, does that mean that the validators that have duplicates will never be requested for download ? 

The question is more about whether this behavior is intended, not if it is correct.

Wondertan: Every validator after splitIdx might be requested in a fallback case when some of the original respond with garbage or do not respond in time. This is the intended behavior, i.e., not relying on low-stakes validators unless we really need to.

fibre/validator/set.go
Wondertan: updated + also removed floats from here

fibre/validator/set.go
nashqueue: Currently this would mean that the bottom 11 validators will never be requested to participate in the normal case. Which i think is fine . The moment we will have better stake distribution we will have less / no duplicates, so can be "unfair" for the time being.

fibre/validator/set.go
nashqueue: Lgtm

fibre/validator/set.go
```diff
@@ -96,6 +96,77 @@ func (s Set) Assign(commitment rsema1d.Commitment, totalRows, originalRows, minR
...
+	}
+
+	validators = make([]*core.Validator, len(s.Validators))
+	copy(validators, s.Validators)
+
+	// find split point where row assignments start overlapping
+	// each validator contributes max(their stake, minStake) where minStake ensures unique decodability
+	totalStake := s.TotalVotingPower()
+	totalDistributedRows := int64(originalRows) * int64(livenessThreshold.Denominator) / int64(livenessThreshold.Numerator)
+	minStake := (int64(minRows)*totalStake + totalDistributedRows - 1) / totalDistributedRows // ceil division
```
walldiss, line 113: Does this algo assume that validators in set are ordered by stake size?

Wondertan: yup

walldiss (approved):

PR #163, opened 2026-01-22
refactor(fibre): BlobID

Wondertan:
## Summary
 Introduces BlobID as a self-describing identifier that combines blob version with commitment, enabling version-aware blob downloads.

## Motivation
The PFF blob on the core chain already contains the version, allowing nodes following the chain to know which version to request. This change enables specifying that version on the download codepath.

Also, the blob version must be known before requesting data—we need to know the expected data format and protocol parameters to properly decode it. Previously, the download endpoint only accepted a commitment, which prevented support for multiple blob versions.

## Changes
  - DownloadShardRequest: Now accepts blob_id (33 bytes: version + commitment) instead of just commitment
  - Server: No longer hardcodes a specific blob version. Both upload and download endpoints dynamically support any blob version via BlobConfigForVersion, completing the versioning story
  - BlobConfig removed from ServerConfig: Server now derives blob configuration from the request's blob version rather than a static config

Depends on #53


fibre/client_server_test.go
```diff
@@ -204,9 +204,9 @@ func TestClientServerUploadDownload(t *testing.T) {
 						return fmt.Errorf("data mismatch for %s: downloaded %d bytes, expected %d bytes",
 							commitment.String(), len(blob.Data()), len(originalData))
 					}
-					if !blob.Commitment().Equals(commitment) {
+					if !blob.ID().Equals(commitment) {
 						return fmt.Errorf("commitment mismatch: got %s, expected %s",
-							blob.Commitment().String(), commitment.String())
+							blob.ID().String(), commitment.String())
```
rootulp: [question] here it looks like we're comparing a blobID and a commitment which I would expect to always fail because the blob ID contains the version and the commitment shouldn't.

If this is really comparing a blobID with another blobID, can we rename variables? If not, likely needs a fix.

rootulp (approved):
Overall LGTM with just one question

fibre/client_server_test.go
Wondertan: Yes, should be renamed

Wondertan: [8904583](https://github.com/celestiaorg/celestia-app-fibre/pull/163/commits/8904583ed6ee5be850678a8b6c0d3f6dca3e2b9a)

rootulp (dismissed):

fibre/blob_id.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// BlobIDSize is the size of a BlobID in bytes.
+// Format: [version (1 byte) | Commitment (32 bytes)]
+const BlobIDSize = 33
+
+// BlobID uniquely identifies a blob by combining version and commitment.
+// The first byte encodes the blob version, followed by 32 bytes of commitment.
+// This makes BlobIDs self-describing, allowing clients to know the blob format before downloading.
+type BlobID []byte
```
walldiss, line 53: Would using struct of two fields be more go native way? like
struct {
  version
  commitment
  }
  
Not requesting changing, just wonder motivation behind usage of byte slice instead of struct

walldiss (dismissed):

fibre/blob_id.go
Wondertan: The same motivation as to why we did `share.Namespace` this way

walldiss (dismissed):

walldiss (approved):

PR #165, opened 2026-01-26
fix(standalone): use SDK genesis provider to handle AppGenesis format

### LINKED ISSUE #164: fix(standalone): use SDK genesis provider to handle AppGenesis format
## Description

The standalone binary fails to start with the error:

```
error reading GenesisDoc at /Users/rootulp/.celestia-app/config/genesis.json: invalid 64-bit integer encoding "1", expected string
```

## Root Cause

The SDK's `AppGenesis` struct serializes `InitialHeight` as a JSON integer (e.g., `"initial_height": 1`), but CometBFT's `GenesisDoc` expects it as a string (e.g., `"initial_height": "1"`).

The standalone binary uses `node.DefaultGenesisDocProviderFunc(cfg)` which reads the genesis file directly as CometBFT's format, causing the type mismatch error.

## Solution

Replace `node.DefaultGenesisDocProviderFunc(cfg)` with a custom genesis provider that:
1. Reads the genesis file using the SDK's `AppGenesisFromFile`
2. Converts it to CometBFT's `GenesisDoc` using `ToGenesisDoc()`

This is the same pattern used by the multiplexer binary in `multiplexer/internal/utils.go:GetGenDocProvider`.

## Acceptance Criteria

- [ ] Standalone binary starts successfully with SDK-generated genesis files
- [ ] Unit test added to prevent regression

rootulp:
## Summary
- Fixes the standalone binary startup failure caused by genesis file format mismatch
- Adds `getGenDocProvider` function that properly converts SDK's AppGenesis to CometBFT's GenesisDoc
- Adds unit tests to prevent regression

## Problem
The standalone binary was failing to start with:
```
error reading GenesisDoc: invalid 64-bit integer encoding "1", expected string
```

This occurred because `node.DefaultGenesisDocProviderFunc` reads the genesis file directly as CometBFT's format, but the SDK's init command generates an `AppGenesis` file where `InitialHeight` is an int64 (JSON integer) rather than a string.

## Solution
Added a custom `getGenDocProvider` function that:
1. Reads the genesis file using the SDK's `AppGenesisFromFile`
2. Converts it to CometBFT's `GenesisDoc` using `ToGenesisDoc()`

This matches the pattern used by the multiplexer binary in `multiplexer/internal/utils.go:GetGenDocProvider`.

## Test plan
- [x] Unit tests added for `getGenDocProvider`
- [x] `go test ./cmd/celestia-appd/...` passes
- [x] `make build-standalone` succeeds
- [x] `./scripts/single-node-fibre.sh` starts successfully (previously failed)

Closes #164

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/celestia-appd/cmd/start_command_standalone.go
+++ b/cmd/celestia-appd/cmd/start_command_standalone.go
@@ -19,6 +19,7 @@ import (
 	"github.com/cometbft/cometbft/proxy"
 	"github.com/cometbft/cometbft/rpc/client/local"
 	coregrpc "github.com/cometbft/cometbft/rpc/grpc"
+	cmttypes "github.com/cometbft/cometbft/types"
 	db "github.com/cosmos/cosmos-db"
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/server"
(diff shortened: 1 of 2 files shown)


Wondertan (approved):
utACK

PR #187, opened 2026-02-17
fix: signatures order when submitting pff

rach-id:
## Overview

This PR fixes an issue with signatures order. When submitting a PFF, we provide the signatures in order, however, if some signatures are missing (given we send the PFF the moment we have 2/3rd), then the validation in the state machine can be invalid given it expects the signatures to be in the order of the validator set.

Note: Claude generated

### DIFF
--- a/fibre/client_upload.go
+++ b/fibre/client_upload.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"sync"
 	"sync/atomic"
 
 	"github.com/celestiaorg/celestia-app-fibre/v6/fibre/validator"
@@ -80,6 +81,9 @@ func (c *Client) Upload(ctx context.Context, ns share.Namespace, blob *Blob) (re
 	requests := makeUploadRequests(shardMap, promise.ToProto(), blob.RLCCoeffs())
 	sigSet := valSet.NewSignatureSet(c.cfg.SafetyThreshold, validatorSignBytes)
 
+	// concurrent-safe map to track validator address → signature for ordering
+	var sigMap sync.Map
+
 	c.log.DebugContext(ctx, "initiating blob upload",
 		"promise_hash", hex.EncodeToString(promiseHash),
 		"promise_height", promise.Height,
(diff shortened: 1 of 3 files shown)


x/fibre/types/msgs.go
```diff
@@ -114,12 +114,6 @@ func (msg *MsgPayForFibre) ValidateBasic() error {
 		return errorsmod.Wrap(sdkerrors.ErrInvalidRequest, "must have at least one validator signature")
 	}
 
-	for i, signature := range msg.ValidatorSignatures {
```
rach-id, line 117: this is removed because we expect empty signatures for validators we couldn't get their signatures in time

fibre/client_upload.go
```diff
@@ -80,6 +81,9 @@ func (c *Client) Upload(ctx context.Context, ns share.Namespace, blob *Blob) (re
 	requests := makeUploadRequests(shardMap, promise.ToProto(), blob.RLCCoeffs())
 	sigSet := valSet.NewSignatureSet(c.cfg.SafetyThreshold, validatorSignBytes)
 
+	// concurrent-safe map to track validator address → signature for ordering
+	var sigMap sync.Map
```
rach-id, line 85: If performance is important here, I could switch to a map + mutex as that's usually faster for heavy writing. However, given this will only have a small number of writes (2/3rd of a validator set of generally 100), I think the performance impact is minimal

rootulp (approved):

Wondertan:
@rach-id, we have an agreement to wait for at least one review by a code owner of the client code, before merging and @rootulp is not a codeowner of fibre client.

fibre/client_upload.go
```diff
@@ -90,31 +94,39 @@ func (c *Client) Upload(ctx context.Context, ns share.Namespace, blob *Blob) (re
...
+	if err = c.uploadShards(ctx, requests, blob, sigSet, &sigMap); err != nil {
 		span.RecordError(err)
 		span.SetStatus(codes.Error, "failed to upload")
 		return result, err
 	}
 
 	// 5) collect signatures
-	sigs, err := sigSet.Signatures()
-	if err != nil {
+	if _, err := sigSet.Signatures(); err != nil {
```
Wondertan, line 104: Didn't that look wrong, that you now completely ignore the values returned?

fibre/client_upload.go
Wondertan, line 85: This is completely unnecessary. There is already `SignatureSet` in the code you changed. Please read the code you changed!

fibre/client_upload.go
```diff
@@ -263,6 +279,7 @@ func (c *Client) uploadShards(
 	requests map[*core.Validator]*types.UploadShardRequest,
 	blob *Blob,
 	sigSet *validator.SignatureSet,
+	sigMap *sync.Map,
```
Wondertan, line 282: We literally pass sigSet here...

fibre/client_upload.go
```diff
@@ -90,31 +94,39 @@ func (c *Client) Upload(ctx context.Context, ns share.Namespace, blob *Blob) (re
...
 	}
 
+	// Build ordered signatures matching the validator set order.
+	// Validators that didn't sign get nil entries.
+	orderedSigs := make([][]byte, len(valSet.Validators))
+	for i, val := range valSet.Validators {
+		if sig, ok := sigMap.Load(val.Address.String()); ok {
+			orderedSigs[i] = sig.([]byte)
+		}
+	}
```
Wondertan, line 117: This should be done in `SignatureSet`. One of the options is making the internal slice a map

fibre/client_upload.go
Wondertan: This comment makes no sense. There is already a mutex in `SignatureSet` that is still used in the code and collects signatures. I will repeat, read the code you change!

Wondertan (commented):
@rach-id, It's clear that the solution is designed without taking existing code into account, meaning you didn't even try to read the code you changed. There is a `SignatureSet` data structure that is designed to hold signatures, which you completely ignored by introducing the `SyncMap` to hold signatures additionaly

fibre/client_upload.go
rootulp: not necessarily. It's possible to check for the presence of an error and not do anything with the values returned.

rootulp:
> @rach-id, we have an agreement to wait for at least one review by a code owner of the client code, before merging and @rootulp is not a codeowner of fibre client.

That's not currently how approvals work on this repo. If you think codeowners should be scoped to portions of the code then please update the CODEOWNERS file with directories and the codeowners for each.

fibre/client_upload.go
Wondertan: The fact that it is possible doesn't make it a justice

Wondertan:
@rootulp, There is an agreement that client code must be reviewed by either Vlad or myself. The fact that its not reflected in the code owners doesn't change it

rootulp:
> @rootulp, There is an agreement that client code must be reviewed by either Vlad or myself. The fact that its not reflected in the code owners doesn't change it

That's an implicit assumption. It's not obvious to new contributors that this "agreement" exists. You could make that assumption explicit by modifying a CODEOWNERS file.

PR #188, opened 2026-02-18
chore: use the signature set for signatures ordering

rach-id:
## Overview

Mainly addresses the feedback in https://github.com/celestiaorg/celestia-app-fibre/pull/187


fibre/validator/signature_set.go
```diff
@@ -46,29 +50,36 @@ func (ss *SignatureSet) Add(val *core.Validator, signature []byte) (bool, error)
...
+// Validators that did not sign have nil entries.
+// Returns [NotEnoughSignaturesError] if voting power threshold is not met.
 // The error contains the partially collected signatures and threshold information.
 func (ss *SignatureSet) Signatures() ([][]byte, error) {
 	ss.mu.Lock()
 	defer ss.mu.Unlock()
 
-	powerNotMet := ss.votingPower < ss.minRequiredVotingPower
-	if powerNotMet {
+	ordered := make([][]byte, len(ss.validators))
```
rach-id, line 67: an alternative would be to do this just at the moment of submission and not here, if that's desired, I can update the PR

rootulp (approved):

fibre/validator/signature_set.go
Wondertan: I think that's fine for the SignatureSet to guarantee order.

fibre/validator/signature_set.go
```diff
@@ -11,13 +11,16 @@ import (
...
+// with nil entries for validators that did not sign.
 type SignatureSet struct {
 	requiredBytesSigned    []byte
 	minRequiredVotingPower int64
+	validators             []*core.Validator
 
 	mu          sync.Mutex
 	votingPower int64
-	signatures  [][]byte
+	signatures  map[string][]byte
```
Wondertan, line 23: We can key with a Validator pointer to avoid allocating a string and copying bytes there.

Wondertan (approved):

PR #192, opened 2026-02-19
chore: reduce the fibre upload log to debug level

rach-id:
## Overview

When you spam fibre, the transactions uploads logs clutter the validator logs. This PR reduces their log level to debug. IMO should be even trace level but it's not implemented. At some point we might want to switch to comet's logger or cosmos-sdk given we implemented trace level there.


Wondertan (approved):

Wondertan:
> At some point we might want to switch to comet's logger or cosmos-sdk given we implemented trace level there

This is what traces are for and we already have tracing coverage.

PR #203, opened 2026-02-23
refactor(fibre)!: independent binary

### LINKED ISSUE #132: fibre/server: graceful shutdown
The core node embedding fibre server currently terminates abruptly, leaving in-process requests unfulfilled and in a partial state. While clients can rely on retries, it is not optimal, and graceful shutdown is nicer. This could be as simple as a waitgroup that waits on Close while preventing newer requests from returning a proper error code.

### LINKED ISSUE #173: Enable fibre server to run in standalone mode
### Summary

Enable the fibre server in celestia-app-fibre to run in standalone mode.

### Problem Definition

Users may want to run standalone so:
- The resources of the fibre server doesn't compete with the consensus node
- Validators want to keep their consensus nodes private (for DoS resistance) while the fibre server is public

### Proposal

To enable the fibre server to run in standalone mode—including on a separate machine—we likely need to do the following:

1. **Remote gRPC Connections**: Ensure validator set, query client, and all chain queries can be performed over the network, not via direct invocation.
2. **PrivValidator Decoupling & Key Management**
   - CLI/config option to specify a validator key path, allowing the fibre server to load or generate validator keys independently.
   - If on a separate machine, should the user copy the keys across? 
   - Support remote signing
3. **Independent Store Directory**
   - CLI/config option for a custom, node-specific storage path, including support for Badger DB or in-memory DB as appropriate.
4. **Standalone gRPC Server**
   - Launch and bind its own gRPC server (with ports configurable via CLI or environment).
   - Register fibre service with this server for remote access.
5. **Integration & Network Tests**

Wondertan:
## Summary

This PR introduces the standalone Fibre binary, separated from `celestia-appd`. Alongside,
several refactorings simplify server construction by replacing manually-injected
dependencies with automatic construction from config parameters.

### Standalone Binary (`fibre/cmd/`)
- **CLI**: Uses Cobra with the simplest possible pattern for horizontal
scalability — new commands and flags can be added without refactoring existing code.
Currently ships two commands: `start` and `version`.
- **Home directory initialization**: On first run, `start` automatically initializes
`FIBRE_HOME` (`~/.celestia-fibre`) with a TOML config file that includes in-file
documentation for every field. Supports both `--home` flag and `FIBRE_HOME` environment
variable, with the env var taking precedence over the default path.
  - The pattern with the `init` command was intentionally avoided.
- **Config precedence**: Implements a 3-tier precedence system (flag > config file >
default), explicitly tested to ensure predictable behavior.
- **Unit testing pattern for commands**: Sets up a reusable test pattern for CLI commands
(config precedence, flag resolution, home-dir initialization), which is usually overlooked.
- **New binary support:** updated goreleaser, Makefile, build scripts, and CI workflows for the new binary.
- [**README:**](https://github.com/celestiaorg/celestia-app-fibre/blob/2cc7ed506e432e1a5cb1937acd458dd5389fdc17/fibre/cmd/README.md) adds readme with simple manual on the binary usage + a little guide on setting up the remote signer
  
### Server Dependency Simplification
- **Before**: Server constructor required callers to manually provide `StateClient`,
`PrivValidator`, and other dependencies.
- **After**:  Only ServerConfig is required. `ServerConfig.Validate()` constructs sensible defaults from config fields. while still allowing overrides for testing or power users.
- **Remote signer (`fibre/sign/`)**: Node operators are expected to use a remote PrivVal
(e.g. tmkms) for signing payment promises. The new `sign` package manages the remote signer
connection and caches the public key on start.
- **gRPC server wrapper (`fibre/grpc/server.go`)**: New transport layer that cleanly
separates gRPC concerns from business logic, with graceful shutdown via `grpc.GracefulStop`.

### Start/Stop Lifecycle
- Follows a proven Start/Stop lifecycle that cleanly separates I/O: `Start` performs all
network connections, background prefetches, and process initialization; `Stop` allows
graceful shutdown of all processes, with context-based force-shutdown for halt scenarios.
- This separation enables predictability and debuggability — callers know exactly when I/O
happens.

### StateClient (`fibre/state_client.go`)
- Introduces `StateClient`, a single interface encapsulating all core/app node dependencies
(chain ID detection, validator set lookups, payment promise verification).
- Concrete implementation in `AppClient` (`fibre/grpc/app_client.go`) uses lazy gRPC
connections — no I/O at construction time, making it safe to use during config validation.
- Prefetches the chain ID on `Start`, simplifying configuration — Fibre now needs only a
single gRPC address to start.
- Simplifies testing: one mock replaces multiple scattered dependency mocks.
- Paves the way for gradual decoupling from the app node.

Closes celestiaorg/celestia-app-fibre#173 
Closes celestiaorg/celestia-app-fibre#132


fibre/cmd/start_cmd.go
```diff
@@ -0,0 +1,73 @@
...
+	"fmt"
+
+	"github.com/celestiaorg/celestia-app-fibre/v6/fibre"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+const (
+	flagListenAddress       = "listen-address"
+	flagStateAddress        = "state-address"
```
cmwaters: I think "consensus-grpc-address" would be more descriptive of what this is than state address. I can imagine users asking what is a "state-address"

fibre/grpc/app_client.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// NewAppClient creates an [AppClient] connected to the given address.
+// The underlying gRPC connection is lazy — no network I/O happens until the first RPC.
+// Call [Start] to auto-detect the chain ID from the node.
+func NewAppClient(addr string) (*AppClient, error) {
+	// grpc.NewClient is lazy — no network I/O happens here, so this is safe outside of Start.
+	conn, err := grpclib.NewClient(
+		addr,
+		grpclib.WithTransportCredentials(insecure.NewCredentials()),
```
cmwaters, line 36: Note: we no longer specify the `MaxMessageSize` but I presume there will be no gRPC calls that exceed 4MB

fibre/sign/remote.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/cometbft/cometbft/crypto"
+	cmtlog "github.com/cometbft/cometbft/libs/log"
+	pvm "github.com/cometbft/cometbft/privval"
+	"github.com/cometbft/cometbft/types"
+)
+
+const (
+	// DefaultListenAddress is the default address for the remote signer listener.
+	DefaultListenAddress = "tcp://127.0.0.1:26658"
```
cmwaters: Would this conflict with the bridge node if it were run on the same machine?

cmwaters (commented):
I notice now that the fibre now talks to the signer over RPC right? What are the changes that this has for node operators? Do we need to document running of the machine.

I think it would be good to do #142 to give node operators visibility into the machine they are running and to be able to set up alerts when something goes wrong - what do you think?

fibre/sign/remote.go
Wondertan: It would. I'll update to non conflicting

fibre/grpc/app_client.go
Wondertan: We didn't specify it before either for app's grpc

fibre/cmd/start_cmd.go
Wondertan: Both flag and config describe what flag is for

Wondertan:
> What are the changes that this has for node operators? 

I don't know exactly. These are more comet-specific details that I would expect ex-core/app team members to know better. 

> Do we need to document running of the machine.

Yes, I presume we will have docs for fibre somewhere on the website. Could be added to the readme somewhere to where fibre server resides too.

Wondertan:
> I think it would be good to do https://github.com/celestiaorg/celestia-app/issues/6808 to give node operators visibility into the machine they are running and to be able to set up alerts when something goes wrong - what do you think?

Yes, this is inevitable.

fibre/state_client.go
```diff
@@ -0,0 +1,30 @@
...
+	"time"
+
+	fibregrpc "github.com/celestiaorg/celestia-app-fibre/v6/fibre/grpc"
+	"github.com/celestiaorg/celestia-app-fibre/v6/fibre/validator"
+	"github.com/celestiaorg/celestia-app-fibre/v6/x/fibre/types"
+)
+
+// StateClient encapsulates everything fibre server depends on to app/core node.
+// The default implementation is [fibregrpc.AppClient].
+type StateClient interface {
```
rach-id: [nit]
I would rename this to `CoreClient` or `AppClient` so that it's more reflective on what state is

fibre/cmd/start_cmd.go
```diff
@@ -0,0 +1,73 @@
...
+	flagSignerListenAddress = "signer-listen-address"
```
rach-id, line 15: this is the privval signer right?

fibre/cmd/start_cmd.go
```diff
@@ -0,0 +1,73 @@
...
+	cmd := &cobra.Command{
+		Use:   "start",
+		Short: "Start the standalone fibre server, initializing home dir and config (first run only)",
+		PreRunE: func(cmd *cobra.Command, _ []string) error {
+			home, err := cmd.Flags().GetString(flagHome)
+			if err != nil {
+				return fmt.Errorf("get %q flag: %w", flagHome, err)
+			}
+
+			if err := initServerConfig(home); err != nil {
```
rach-id, line 33: I saw the comment that it's intentional to `init` inside the `start` command, but it's been the standard for all our binaries. We should document this clearly in the docs for operators

fibre/sign/remote.go
Wondertan: Fixed

Wondertan:
@cmwaters, added [readme](https://github.com/celestiaorg/celestia-app-fibre/blob/2cc7ed506e432e1a5cb1937acd458dd5389fdc17/fibre/cmd/README.md) for the binary with the instructions on setting up the tmkms. The only issue is that it doesn't include the migration guide that some operators might need. Irrc, we already migrated off Hocrux, but I have no clue what the current role model setup is for them to migrate from. Also, bringing here the BLS angle, committing to BLS would allow us to avoid the migration hassle, as well as backwards compatibility(actually the worst) if we skip tmkms for the mainnet and do BLS directly.

fibre/grpc/app_client.go
```diff
@@ -0,0 +1,93 @@
...
+	coregrpc "github.com/cometbft/cometbft/rpc/grpc"
+	tmservice "github.com/cosmos/cosmos-sdk/client/grpc/cmtservice"
+	grpclib "google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+// AppClient manages a gRPC client connection to a celestia-app node
+// and provides the query methods needed by the Fibre server.
+type AppClient struct {
+	*SetGetter
```
rach-id, line 22: SetGetter is very unintuitive, I would propose `ValidatorSetGetter` or something easier to understand

fibre/grpc/app_client.go
```diff
@@ -0,0 +1,93 @@
...
+	queryClient types.QueryClient
+
+	chainID string // resolved on Start
+}
+
+// NewAppClient creates an [AppClient] connected to the given address.
+// The underlying gRPC connection is lazy — no network I/O happens until the first RPC.
+// Call [Start] to auto-detect the chain ID from the node.
+func NewAppClient(addr string) (*AppClient, error) {
+	// grpc.NewClient is lazy — no network I/O happens here, so this is safe outside of Start.
```
rach-id: [nit]
seems redundant with the method comment
```suggestion
```

fibre/sign/remote.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		_ = endpoint.Stop()
+		return nil, fmt.Errorf("creating privval signer client: %w", err)
+	}
+
+	retrySigner := pvm.NewRetrySignerClient(client, retries, retryTimeout)
+	pubKey, err := retrySigner.GetPubKey()
+	if err != nil {
+		_ = errors.Join(client.Close(), endpoint.Stop())
+		return nil, fmt.Errorf("getting public key from remote signer: %w", err)
```
rach-id, line 54: probably also log the error from closing the client/endpoint

fibre/server_config.go
```diff
@@ -0,0 +1,150 @@
...
+		StoreConfig:         DefaultStoreConfig(),
+		LivenessThreshold:   p.LivenessThreshold,
+		MinRowsPerValidator: p.MinRowsPerValidator(),
+		MaxMessageSize:      p.MaxMessageSize(),
+	}
+	return cfg
+}
+
+// Validate validates the ServerConfig and sets default values for unset fields.
+func (cfg *ServerConfig) Validate() error {
```
rach-id, line 87: [nit]
given this initialises also some fields, can its name be renamed to reflect that?

fibre/grpc/app_client.go
rach-id: This seems to be only for querying the validator set, so it's fine

fibre/cmd/start_cmd.go
rach-id: I second Callum, `state` is ambiguous and could use a renaming

fibre/cmd/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Usage
+
+### Start
+
+```sh
+fibre start
+```
+
+On first run, initializes `~/.celestia-fibre` with a default TOML config.
```
rach-id, line 21: what if the first ran has some flags? will those be used to initialise the config file?

fibre/cmd/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Version
+
+```sh
+fibre version
+```
+
+## Config
+
+The config file is at `$FIBRE_HOME/config.toml` (default `~/.celestia-fibre/config.toml`).
```
rach-id: is this the same as `server_config.toml`? if not, it would be good to document both

fibre/cmd/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+Config precedence: **flag > config file > default**.
+
+## Remote Signer
+
+Fibre uses a remote PrivVal signer (e.g. [tmkms](https://github.com/iqlusioninc/tmkms)) to sign payment promises. The signer **dials into** the fibre server's listener address.
```
rach-id, line 55: probably worth noting that this can also use the local signer and I assume it's the default currently no?

rach-id (commented):
Overall LGTM, left some comments to understand more.

Additionally, this PR breaks the integration of fibre in talis. We should create an issue to update talis after this is merged.

Also, this PR seems to be breaking, the title should include !

rach-id:
> The only issue is that it doesn't include the migration guide that some operators might need

I don't think we should provide that here. We could do it separately but it's the operator's responsability to migrate.

> we already migrated off Hocrux

It's still used by many validators and they're wondering wether a latency of ~10ms is fine for fibre especially at first. 

Also, if they don't have the seed they used to generate the horcrux keys shards, then there is no way to go to tmkms and will need to create a new validator.

fibre/state_client.go
Wondertan: The thing is, it's both core and app, thus I named it as something what encompasses them both

fibre/cmd/start_cmd.go
Wondertan: The standard has been a separate `init` cmd, and I voided this here.

fibre/cmd/start_cmd.go
Wondertan: yup

fibre/server_config.go
Wondertan: You can technically sqeeuze in addtional meaning into what Validate is, but I am open to suggestions

fibre/cmd/README.md
Wondertan: nope

Wondertan: init always uses the default defaults, but we can change that with a single loc change if we want.

Wondertan:
> I don't think we should provide that here. We could do it separately but it's the operator's responsability to migrate.

Well, it's in our interest to make sure validators don't fuck up the signing process. Although, no strong opinions on this, I am not dealing with them anyway.

fibre/cmd/README.md
Wondertan: We used the whatever signer was configured on coreapp binary. Now it's only the remote. The local is not supported in this PR, but I am all ears on some good arguments on why we should provide support for it.

fibre/grpc/app_client.go
Wondertan: It's used for querying everything from core/app not only val sets

fibre/cmd/start_cmd.go
Wondertan: And yeah, its documented

fibre/grpc/app_client.go
Wondertan: The main thing is how interface is named, which is `validator.SetGetter`

fibre/cmd/README.md
Wondertan: its `server_config.toml`, fixed now

fibre/cmd/start_cmd.go
Wondertan: Called it `app-grpc-addresss` as it's technically an `app`s grpc server.

fibre/grpc/app_client.go
cmwaters: But I don't think anything here will exceed 4MB right

Wondertan: I can't tell for sure, but I would expect validator sets, validator registry, and payment promise validation requests to be less then that

cmwaters:
I can create a follow up issue to add information on migrating in the release notes. 

As I understand this, all operators will need to run a remote signer (at least for their fibre server).

cmwaters:
@rach-id created https://github.com/celestiaorg/celestia-core/issues/2829 for updating talis

fibre/state_client.go
```diff
@@ -0,0 +1,30 @@
...
+	// SetGetter is embedded to provide validator set lookups.
+	validator.SetGetter
+
+	// ChainID returns the chain ID of the state machine.
+	ChainID() string
+	// Verify verifies a payment promise against on-chain state
+	// and returns the expiration time.
+	Verify(context.Context, *types.PaymentPromise) (time.Time, error)
```
evan-forbes: nice interface

we might want to return a new struct here so that we can change verification in a non-breaking way

in the future we could move almost entirely to the server and only query once a day per account

fibre/cmd/start_cmd.go
```diff
@@ -0,0 +1,73 @@
...
+// RunE after config resolution; passing it as a parameter keeps the command
+// testable without global state.
+func newStartCmd(start func(context.Context, fibre.ServerConfig) error) *cobra.Command {
+	cfg := fibre.DefaultServerConfig()
+
+	cmd := &cobra.Command{
+		Use:   "start",
+		Short: "Start the standalone fibre server, initializing home dir and config (first run only)",
+		PreRunE: func(cmd *cobra.Command, _ []string) error {
+			home, err := cmd.Flags().GetString(flagHome)
```
evan-forbes, line 28: is the home path what is passed to the store? if so, do we want that and not like a fibre/data/ directory?

https://github.com/celestiaorg/celestia-app-fibre/blob/feature/fibre/fibre/store.go?plain=1#L56

scripts/single-node-fibre.sh
```diff
@@ -119,11 +140,22 @@ startCelestiaApp() {
   echo "celestia-appd started with PID: ${CELESTIA_APP_PID}"
 }
 
-deleteCelestiaAppHome
+startFibre() {
+  echo "Starting fibre in background..."
+  fibre start \
+    --home "${FIBRE_HOME}" \
+    --app-grpc-address "${APP_GRPC_ADDR}" &
```
evan-forbes, line 147: is the default being used for the signer here?

> "tcp://127.0.0.1:26659"

and are we relying on tmkms or something else being setup to run this script? 

if so, this is definitely acceptable for now as we can flup

evan-forbes (dismissed):
no blocking comments

opened an issue to decide on the kms stuff https://linear.app/celestia/issue/DA-1102/decide-on-key-management-for-fibre-now-that-there-is-not-a-single . what does the single-node-fibre script do for a kms? are we reliant on something else being spun up?

fibre/state_client.go
Wondertan: Appreciate the compliment. This interface should help us, no matter which direction we take with the fibre <-> core/app relation.

Wondertan: > we might want to return a new struct here so that we can change verification in a non-breaking way

I am not opposed to future-proving this a little bit, but would like to understand what else do you expect Verify to return?

fibre/cmd/start_cmd.go
Wondertan: Badger creates a subfolder automatically. We could layer additional directory there, but I avoided that.

scripts/single-node-fibre.sh
Wondertan: Yes, it uses the default here. It's a good catch, actually, because I didn't think about tmkms within a single script setup here and it may not work

Wondertan: Ok, I see https://github.com/celestiaorg/celestia-core/issues/2838, so keeping it a flup

fibre/state_client.go
evan-forbes: the current escrow balance could be returned

so that we only need to call verify when the fibre server doesn't already have a balance cached

Wondertan: Hmm, what's the idea behind fibre level caching for balances? I am a bit concerned about spreading persistent states between the app and the fibre. It may give some performance boost short term, but has the potential to make a mess harder to untangle long term

evan-forbes: >  I am a bit concerned about spreading persistent states between the app and the fibre

fair yeah that seems like the tradeoff. this is to make promise verification not require pinging the the statemachine each time. It definitely adds things to the fibre server, however I think its a nice option in the future if we add some form of relay where users are submitting potentially many many times in a single day.

this can also just be added later if we need it. Its not that big of a deal to break a mostly internal interface via switch to a struct that has both

Wondertan: It's not an argument against future-proofing, but more a concern about the state optimization. Anyway, I future proved this with a struct which also needed change to pkg structure to solve cyclical dependency.

Wondertan:
Added more comprehensive logging and the ability to control verbosity within the binary. Same for otel tracing, pyroscope and pprof. 

They are all independent commits, which could become their own PR, but considering the moving, I am not sure if creating new PRs is worth it.

evan-forbes (dismissed):
the rest of the introspection is added in the expected way so 👍 LGTM :shipit:

evan-forbes (approved):
next time we're going straight to out of process separate binary

Wondertan:
> next time we're going straight to out of process separate binary

Next time? Isnt this what we are doing?

cmwaters:
> Next time? Isnt this what we are doing?

I think he means instead of starting in the same binary and then splitting it out we'd just start with a separate binary

PR #214, opened 2026-02-24
refactor(fibre)!: simplify Client construction

Wondertan:
## Summary
Simplifies the `Client` construction by building on ideas from #203 and updates existing tools and tests.

### Constructor Simplification
  - **Before**: `NewClient(txClient, keyring, valSetGetter, hostRegistry, cfg)` — callers had
  to manually create and wire `TxClient`, `SetGetter`, and `HostRegistry` from raw gRPC
  connections.
  - **After**: `NewClient(keyring, cfg)` — dependencies are constructed automatically via
  `StateClientFn`, which defaults to creating an `AppClient` from `cfg.StateAddress`.
  - Removes `ChainID` from `ClientConfig` — now auto-detected from the node during `Start`.
  - Removes `--chain-id` flag from both CLI tools (`fibre-txsim`, `submit-fibre-blob`).

### Client Start/Stop Lifecycle
- Adds `Client.Start(ctx)` which initializes the underlying `StateClient` (chain ID
detection, host registry prefetch).
- `Upload` and `Download` return an error if called before `Start`, preventing misuse.
- Mirrors the server's Start/Stop pattern for consistency across the codebase.

### StateClient: Shared Server/Client Abstraction
- Embeds `validator.HostRegistry` into the `StateClient` interface alongside the existing
`validator.SetGetter`, unifying all app-node dependencies under one abstraction.
- `AppClient` now creates and owns both `SetGetter` and `HostRegistry` from the same gRPC
connection.
- `AppClient.Start` runs chain ID detection and host registry prefetch in parallel.

### `internal` package
* Since most of the dependencies for both server and client are now constructed internally, there is no point in exporting them, so we move them to the `internal` pkg. 
* This mainly allows breaking the internal API without worries about breaking any external packages.
* The rule of thumb is to add internal things in the respective package and only expose them once actually needed.

Depends on #203


fibre/client_put.go
```diff
@@ -29,8 +30,12 @@ type PutResult struct {
...
 // Put uploads given data to the Fibre network.
 // It encodes the data into a [Blob], calls [Client.Upload] to upload it,
-// and submits a MsgPayForFibre transaction.
-func (c *Client) Put(ctx context.Context, ns share.Namespace, data []byte) (result PutResult, err error) {
+// and submits a MsgPayForFibre transaction using the provided [user.TxClient].
+//
+// TODO(@Wondertan): This does not belong here. Fibre protocol in it's core doesn't need to know about transactions.
+// Furthermore, this function cannot be generalized for all the cases with fee grants, multiple key managements, etc.
+// And users are strongly advised to use [fibre.Upload] with custom TX submission logic instead, ideally batching multiple blobs in a single PFF.
+func Put(ctx context.Context, c *Client, txClient *user.TxClient, ns share.Namespace, data []byte) (result PutResult, err error) {
```
evan-forbes, line 38: hmmmm

evan-forbes (dismissed):

fibre/client_put.go
Wondertan: Yes, this might be forward-looking on my side, but given how nodes send txs, it is much easier to construct and send PFF yourself than to use Put. Node or any other tx submitter needs to handle feegranter, multiple key selection, fee calculation, and prioritization, which are all not concerns of the fibre client. Beyond, not everyone needs to block on the tx finalization, just like some will prefer batching multiple uploads into a single PFF once we add this

evan-forbes (approved):

PR #217, opened 2026-03-01
feat: monitor the server's CPU/RAM usage during talis experiments

rach-id:
## Overview

This is beneficial to know the resources usage during experiments and plot them


mcrakhman (commented):
Really like the idea to do the resource monitoring, I think it is super useful for debugging etc!

The only thing I am not sure I like the approach with shell scripts here, because it is very hard to read/debug and maintain.

I understand that copying another binary to vals may sound cumbersome, but at the same time, we can just do it once (have single binary) for all helper commands and then have go code which is easier to read/test etc.

and have something like `utils resource-monitor --...`

rach-id:
> it is very hard to read/debug and maintain.

I agree with you. I didn't read it anyway, it was claude generated. And I don't think anyone should maintain it, they should just ask claude to modify it if needed or generate a new one.

This functionality helped me have an idea on the resources usage in Talis. But I don't think it's worth spending time writing clean code for it. That's why I opened this PR. 

If you feel strongly against it, let's close the PR and write an issue and we can work on it afterwards.

mcrakhman:
> > it is very hard to read/debug and maintain.
> 
> I agree with you. I didn't read it anyway, it was claude generated. And I don't think anyone should maintain it, they should just ask claude to modify it if needed or generate a new one.
> 
> This functionality helped me have an idea on the resources usage in Talis. But I don't think it's worth spending time writing clean code for it. That's why I opened this PR.
> 
> If you feel strongly against it, let's close the PR and write an issue and we can work on it afterwards.

I also agree with you in a way, though I think Claude can generate the go thing easily, and then we would have an extensible binary for things like this etc. But maybe if other reviewers are ok with the shell script approach, I won't insist on my stuff

rootulp (approved):

PR #232, opened 2026-03-04
docs: add an initial implementation plan for the fibre server local cache

rach-id:
## Overview

This first part only included the design decisions for this. Once we agree on them, I can generate a detailed implementation plan and then go for the implementation

### DIFF
--- a/fibre/plan.md
+++ b/fibre/plan.md
@@ -95,7 +95,7 @@ A payment promise accepted by the fibre server can be rejected on-chain via `Msg
 **Answer**: No. Same UX concern. The cache solves the problem without changing economic parameters.
 
 ### DD11: Can we skip `ValidatePaymentPromise` entirely and do all checks locally?
-**Answer**: See DD2 for the decision. The long-term direction is to move as much validation logic as possible to run locally, with event-driven updates for chain state changes.
+**Answer**: See DD2 for the decision. It's partially possible to do given the details in DD2.
 
 ### DD12: What about deposits (balance increases) happening after cache was seeded?
 **Answer**: The cache won't see new deposits until TTL expiry or zero-balance re-query. When `available` hits zero, the next promise triggers a re-fetch, which picks up any deposits.
--- a/fibre/plan.md
+++ b/fibre/plan.md
@@ -42,7 +42,9 @@ Checks 1-4 only need `blockTime`, `blockHeight`, and `params` — all refreshabl
 **Answer**: On cache miss, after the `ValidatePaymentPromise` gRPC call succeeds, make a separate `EscrowAccount` gRPC query to get the current `AvailableBalance`. The `EscrowAccount` request already exists and returns the balance.
 
 ### DD4: What is the cache TTL? (addresses parameter change fragility)
-**Answer**: `TTL = WithdrawalDelay - PaymentPromiseTimeout` (default: 24h - 1h = 23h). Derived dynamically from chain params, never hardcoded. Params can be refreshed once per block, or by listening for the existing `EventUpdateFibreParams` event (already emitted in `msg_server.go:289` when `UpdateFibreParams` is called — includes the full `Params` struct). No new events need to be added.
+**Answer**: `TTL = WithdrawalDelay - PaymentPromiseTimeout` (default: 24h - 1h = 23h). Derived dynamically from chain params, never hardcoded. Params are refreshed once per block by listening for the existing `EventUpdateFibreParams` event (already emitted in `msg_server.go:289` when `UpdateFibreParams` is called — includes the full `Params` struct). No new events need to be added.
+
+To avoid a race condition between a parameter change and promise processing, the cache must **lock on every new block**, update `blockTime`, `blockHeight`, params (and any other per-block state), and only then unlock to resume processing promises. This ensures no promise is validated against stale state mid-block transition.
 
 ### DD5: When does the cache re-query the state machine?
 Three triggers:
--- a/fibre/plan.md
+++ b/fibre/plan.md
@@ -83,7 +83,7 @@ A payment promise accepted by the fibre server can be rejected on-chain via `Msg
 **Additional rejection for `MsgPaymentPromiseTimeout`** (checks 9-11 are skipped):
 18. Not yet timed out: `blockTime < expirationTime` (the inverse of check 9)
 
-**Answer**: So, if we implement the local validation of the payment promises. We can be sure that any payment promise that's submitted onchain and has enough signatures is valid.
+**Answer**: So, if we implement the local validation of the payment promises. We can be sure that any payment promise that's submitted onchain and has enough signatures is valid. And we don't need to worry about this case. For the expired promises, that's handled in another issue.
 
 ### DD8: Thread-safety approach?
 **Answer**: Single `sync.Mutex` on the cache. On cache hit: lock, deduct, unlock. On cache miss: lock to check, unlock, call gRPC (no lock held), lock to populate and deduct, unlock. Duplicate fetches on concurrent misses are harmless (both get the same block's state).
--- a/fibre/plan.md
+++ b/fibre/plan.md
@@ -98,6 +98,11 @@ A payment promise accepted by the fibre server can be rejected on-chain via `Msg
 **Answer**: See DD2 for the decision. The long-term direction is to move as much validation logic as possible to run locally, with event-driven updates for chain state changes.
 
 ### DD12: What about deposits (balance increases) happening after cache was seeded?
-**Answer**: The cache won't see new deposits until TTL expiry or zero-balance re-query. When `available` hits zero, the next promise triggers a re-fetch, which picks up any deposits. Documented as a known limitation.
+**Answer**: The cache won't see new deposits until TTL expiry or zero-balance re-query. When `available` hits zero, the next promise triggers a re-fetch, which picks up any deposits.
+
+**Note**: If we adopt the per-block refresh pattern from DD4, we could also listen for `EventDepositToEscrow` (already emitted in `msg_server.go:64`) and update the cached balance for the affected account immediately. This would eliminate this limitation entirely.
+
+### DD13: Cache size and eviction
+**Answer**: The `entries` map grows unbounded as unique accounts submit promises. Entries should be evicted after TTL expiry to prevent unbounded memory growth. A simple approach: during the per-block update (DD4), iterate entries and remove any where `time.Since(entry.fetchedAt) >= ttl`. This piggybacks on the existing per-block lock so no additional synchronization is needed.
 
 ---
--- a/fibre/plan.md
+++ b/fibre/plan.md
@@ -20,11 +20,29 @@
 
 **Proposal**: We could make many validations local. For example, the timestamp could be checked against the block time. We could throw events on params/governance changes and listen for those events so that when they happen, we update immediately. The goal would be to have as much of `ValidatePaymentPromise` logic running locally as possible, reducing reliance on the gRPC round-trip for each promise. This could be a follow-up.
 
+**Full list of validations in `ValidatePaymentPromise`** (from `keeper.go:313-374`):
+
+1. **Timestamp not too old**: `creationTime > blockTime - WithdrawalDelay`. Needs `blockTime`, `params.WithdrawalDelay`. Could be local — cache block time per block, cache params via `EventUpdateFibreParams`.
+
+2. **Not expired**: `blockTime < creationTime + PaymentPromiseTimeout`. Needs `blockTime`, `params.PaymentPromiseTimeout`. Could be local — same as above.
+
+3. **Height not too far in past**: `currentHeight - promiseHeight <= PaymentPromiseHeightWindow`. Needs `blockHeight`, `params.PaymentPromiseHeightWindow`. Could be local — cache block height per block.
+
+4. **Height not too far in future**: `promiseHeight <= currentHeight + 1`. Needs `blockHeight`. Could be local — same.
+
+5. **Not already processed**: `IsPaymentPromiseProcessed()`. Needs processed payments KV store. Harder — need to track which promises have been included on-chain. Could read blocks or listen for `EventPayForFibre`/`EventPaymentPromiseTimeout` events.
+
+6. **Escrow account exists**. Needs escrow accounts KV store. Could be cached, seeded on first query, invalidated on deposits/withdrawals.
+
+7. **Sufficient available balance**: `AvailableBalance >= cost`. Needs escrow account balance. This is exactly what the escrow cache solves.
+
+Checks 1-4 only need `blockTime`, `blockHeight`, and `params` — all refreshable once per block. Check 5 is the hardest to make local. Checks 6-7 are what the escrow cache addresses. The proposal is to query this information once per block instead of once per-payment promise.
+
 ### DD3: How does the cache get seeded with the on-chain balance?
 **Answer**: On cache miss, after the `ValidatePaymentPromise` gRPC call succeeds, make a separate `EscrowAccount` gRPC query to get the current `AvailableBalance`. The `EscrowAccount` request already exists and returns the balance.
 
 ### DD4: What is the cache TTL? (addresses parameter change fragility)
-**Answer**: `TTL = WithdrawalDelay - PaymentPromiseTimeout` (default: 24h - 1h = 23h). Derived dynamically from chain params, never hardcoded. Params themselves are cached with a 10-minute TTL and re-fetched via `Params` gRPC query. Additionally, emit an event on TTL/param changes and listen for them so the cache can update immediately rather than waiting for the polling interval.
+**Answer**: `TTL = WithdrawalDelay - PaymentPromiseTimeout` (default: 24h - 1h = 23h). Derived dynamically from chain params, never hardcoded. Params can be refreshed once per block, or by listening for the existing `EventUpdateFibreParams` event (already emitted in `msg_server.go:289` when `UpdateFibreParams` is called — includes the full `Params` struct). No new events need to be added.
 
 ### DD5: When does the cache re-query the state machine?
 Three triggers:
@@ -36,7 +54,34 @@ Three triggers:
 **Answer**: All accepted payment promises are deducted from the cache regardless of whether they are ultimately included in a block. When the cache balance is renewed (on TTL expiry or zero-balance re-query), any pending (not-yet-included) promises must still be deducted from the refreshed balance. One approach is to read committed blocks to identify which payment promises were actually included on-chain, and maintain a set of pending promises whose deductions carry over across cache refreshes. Without this, malicious actors could benefit from free DA by submitting promises that pass the cache check but are never included on-chain, and then having the cache "forget" those deductions on refresh.
 
 ### DD7: How are promises that were not validated on-chain handled?
-**Answer**: TBD.
+
+A payment promise accepted by the fibre server can be rejected on-chain via `MsgPayForFibre` or `MsgPaymentPromiseTimeout` for any of the following reasons:
+
+**Stateless rejections** (from `payment_promise.go:121-171`):
+1. Invalid signer key (not 33-byte secp256k1 public key)
+2. Empty or oversized chain ID
+3. Zero upload size
+4. Zero creation timestamp
+5. Invalid signature length (not 64 bytes)
+6. Zero height
+7. Signature verification fails
+
+**Stateful rejections for `MsgPayForFibre`** (from `keeper.go:313-374`, `msg_server.go:124-199`):
+8. Timestamp too old: `creationTime <= blockTime - WithdrawalDelay`
+9. Promise expired: `blockTime >= creationTime + PaymentPromiseTimeout`
+10. Height too far in past: `currentHeight - promiseHeight > PaymentPromiseHeightWindow`
+11. Height too far in future: `promiseHeight > currentHeight + 1`
+12. Already processed (promise hash in processed payments store)
+13. Escrow account not found
+14. Insufficient balance (`Balance < paymentAmount`)
+15. Insufficient available balance (`AvailableBalance < paymentAmount`)
+16. Invalid validator sign bytes
+17. Validator signature validation fails (can't get historical validator set, invalid signatures, or < 2/3 voting power)
+
+**Additional rejection for `MsgPaymentPromiseTimeout`** (checks 9-11 are skipped):
+18. Not yet timed out: `blockTime < expirationTime` (the inverse of check 9)
+
+**Answer**: So, if we implement the local validation of the payment promises. We can be sure that any payment promise that's submitted onchain and has enough signatures is valid.
 
 ### DD8: Thread-safety approach?
 **Answer**: Single `sync.Mutex` on the cache. On cache hit: lock, deduct, unlock. On cache miss: lock to check, unlock, call gRPC (no lock held), lock to populate and deduct, unlock. Duplicate fetches on concurrent misses are harmless (both get the same block's state).
--- a/fibre/plan.md
+++ b/fibre/plan.md
@@ -1,0 +1,56 @@
+# Local Escrow Balance Cache for Fibre Server
+
+## Context
+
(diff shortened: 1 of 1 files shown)


fibre/plan.md
```diff
@@ -0,0 +1,56 @@
...
+
+**Proposal**: We could make many validations local. For example, the timestamp could be checked against the block time. We could throw events on params/governance changes and listen for those events so that when they happen, we update immediately. The goal would be to have as much of `ValidatePaymentPromise` logic running locally as possible, reducing reliance on the gRPC round-trip for each promise. This could be a follow-up.
+
+### DD3: How does the cache get seeded with the on-chain balance?
+**Answer**: On cache miss, after the `ValidatePaymentPromise` gRPC call succeeds, make a separate `EscrowAccount` gRPC query to get the current `AvailableBalance`. The `EscrowAccount` request already exists and returns the balance.
+
+### DD4: What is the cache TTL? (addresses parameter change fragility)
+**Answer**: `TTL = WithdrawalDelay - PaymentPromiseTimeout` (default: 24h - 1h = 23h). Derived dynamically from chain params, never hardcoded. Params themselves are cached with a 10-minute TTL and re-fetched via `Params` gRPC query. Additionally, emit an event on TTL/param changes and listen for them so the cache can update immediately rather than waiting for the polling interval.
```
rach-id: The TTL could be updated every 10min. But we could also use the events to see when the params change

fibre/plan.md
```diff
@@ -0,0 +1,56 @@
...
+Three triggers:
+1. **No entry** for this account (first promise ever).
+2. **TTL expired** (`time.Since(entry.fetchedAt) >= ttl`).
+3. **Cached balance is zero** — re-query to pick up deposits that happened after the last fetch.
+
+### DD6: How is inter-block synchronization handled?
+**Answer**: All accepted payment promises are deducted from the cache regardless of whether they are ultimately included in a block. When the cache balance is renewed (on TTL expiry or zero-balance re-query), any pending (not-yet-included) promises must still be deducted from the refreshed balance. One approach is to read committed blocks to identify which payment promises were actually included on-chain, and maintain a set of pending promises whose deductions carry over across cache refreshes. Without this, malicious actors could benefit from free DA by submitting promises that pass the cache check but are never included on-chain, and then having the cache "forget" those deductions on refresh.
+
+### DD7: How are promises that were not validated on-chain handled?
+**Answer**: TBD.
```
rach-id: This is basically the expired payment promisses case and the payment promises that failed validation onchain. Still thinking about it

fibre/plan.md
rach-id: We could also go for checking these only once per block instead of with every fibre blob

fibre/plan.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Design Decisions
+
+### DD1: Where does the cache live?
+**Answer**: In the fibre server (`fibre/` package), as a field on `Server`. The state machine remains the source of truth; the cache is a local optimization.
+
+### DD2: Do we still call `ValidatePaymentPromise` gRPC?
+**Answer**: Yes. The gRPC call handles non-balance stateful checks (timestamp freshness, expiration, height window, not-already-processed) that require chain state. The cache **only** replaces the balance sufficiency check. The gRPC call happens first; the local balance deduction happens after it succeeds.
+
+**Proposal**: We could make many validations local. For example, the timestamp could be checked against the block time. We could throw events on params/governance changes and listen for those events so that when they happen, we update immediately. The goal would be to have as much of `ValidatePaymentPromise` logic running locally as possible, reducing reliance on the gRPC round-trip for each promise. This could be a follow-up.
```
rach-id, line 21: We could also go for checking these only once per block instead of with every fibre blob. This could be a follow up. 

The list of validations and how it can be made local is in the next line

fibre/plan.md
```diff
@@ -0,0 +1,108 @@
+# Local Escrow Balance Cache for Fibre Server
+
+## Context
+
+**Problem**: When a validator receives a `PaymentPromise` via `UploadShard`, it calls `ValidatePaymentPromise` on the state machine (loopback gRPC). The state machine checks `AvailableBalance >= cost`, but this balance is only updated when a `MsgPayForFibre` is included in a block. If the same account sends N promises within a block, all N see the same stale balance and are accepted — enabling over-commitment of escrow funds.
+
+**Goal**: Add a per-account balance cache in the fibre server that tracks local deductions across accepted promises, preventing over-acceptance within and across blocks. The 24h `WithdrawalDelay` guarantees that a queried balance is valid for ~23h (funds can't leave escrow faster than the delay), so the cache can avoid re-querying the state machine on every promise.
```
rootulp, line 7: [question] why ~23 hours and not 24 hours?


```suggestion
**Goal**: Add a per-account balance cache in the fibre server that tracks local deductions across accepted promises, preventing over-acceptance within and across blocks. The 24h `WithdrawalDelay` guarantees that a queried balance is valid for 24h (funds can't leave escrow faster than the delay), so the cache can avoid re-querying the state machine on every promise.
```

fibre/plan.md
```diff
@@ -0,0 +1,108 @@
...
+
+**Ref**: [GitHub Issue #27](https://github.com/celestiaorg/celestia-app-fibre/issues/27)
+
+---
+
+## Design Decisions
+
+### DD1: Where does the cache live?
+**Answer**: In the fibre server (`fibre/` package), as a field on `Server`. The state machine remains the source of truth; the cache is a local optimization.
```
rootulp, line 16: If the state machine is the source of truth, do we already have tests that confirm that the state machine won't include 2 PFFs in the same block if the 2 PFFs double spend from an escrow account?

If not, do we need a FLUP to add that test?

fibre/plan.md
rootulp: Have we identified that a GRPC request to validate the payment promise is a bottleneck? I'm in favor of not pursuing any of these optimizations (i.e. skipping some portion of validate payment promise via a local check) unless they are strictly required for some KPI.

fibre/plan.md
```diff
@@ -0,0 +1,108 @@
...
+**Answer**: Single `sync.Mutex` on the cache. On cache hit: lock, deduct, unlock. On cache miss: lock to check, unlock, call gRPC (no lock held), lock to populate and deduct, unlock. Duplicate fetches on concurrent misses are harmless (both get the same block's state).
+
+### DD9: Should we limit one promise per account per height?
+**Answer**: No. This hurts UX for power users submitting multiple blobs. The balance cache is sufficient to prevent over-commitment without artificial rate limits.
+
+### DD10: Should we require dramatically higher escrow minimums?
+**Answer**: No. Same UX concern. The cache solves the problem without changing economic parameters.
+
+### DD11: Can we skip `ValidatePaymentPromise` entirely and do all checks locally?
+**Answer**: See DD2 for the decision. The long-term direction is to move as much validation logic as possible to run locally, with event-driven updates for chain state changes.
```
rootulp: [question] why is this the long-term direction? One GRPC request per `ValidatePaymentPromise` _seems_ fine. Do we have data that this is slowing down the total Fibre throughput?

rootulp (dismissed):
Zooming out, are there any other alternatives to solving the problem of allowing multiple PFFs in the same block?

The cache (this PR) seems like one potentially fragile approach. I'm wondering if there are any alternative approaches.

fibre/plan.md
rach-id: > Have we identified that a GRPC request to validate the payment promise is a bottleneck?

I don't think that's the case but I didn't run any experiments to see this.

This suggestion goes in line with reducing relying on the state by keeping a temporary local state for verification. But I 100% agree, the less duplicate state the better and I would prefer to keep the local solely for balance verification

fibre/plan.md
rach-id: It's specified in: 
https://github.com/celestiaorg/celestia-app-fibre/blob/2055da64e21609df0fb15f3e3d2976f45c09e92a/fibre/plan.md?plain=1#L44-L45

Claude answer:
```
**Why 23h and not 24h?** Assume I have 10 TIA and the cache is updated every 24h. The cache is updated now and sees 10 TIA. A few seconds later, I start a withdrawal. For the next 24h, the cache still thinks I have 10 TIA. Now assume I create a payment promise between hour 23 and hour 24. The cache validates it — I have 10 TIA, looks good. The validator serves me the DA. But that promise can be submitted on-chain up to 1h after creation (PaymentPromiseTimeout), so it could land on-chain after hour 24. By then, the withdrawal has completed and my escrow balance is 0. The `MsgPayForFibre` fails — but I already got the DA for free.

With a 23h TTL, the cache expires before this window opens. Any promise created before hour 23 must be submitted on-chain before hour 24 (23h + 1h PaymentPromiseTimeout), which is before the withdrawal completes. At hour 23, the cache refreshes, sees `AvailableBalance=0`, and rejects further promises.
```

Prompt (my english):
Assume I have 10 tia and the cache is updated every 24hrs. Then, if the cache is updated now, and after a few seconds, I start the withdrawal. Then for the next 24hrs, the cache thinks I have 10tia.
Assume I create a payment promise between 23hr-24hrs. Then, I would validate that PP from the cache. Thing is, that PP can be submitted after 23hr + 1hr, and will fail given the escrow account has 0 balance. But the problem is that I would have benefited from fibre DA without paying for it.

fibre/plan.md
rach-id: > do we already have tests that confirm that the state machine won't include 2 PFFs in the same block if the 2 PFFs double spend from an escrow account?

If the balance in the escrow account is the same as the cost for that PFF, then the first will succeeded, the second will fail. However, the user would have benefitted from fibre for 2 PFFs and only paid for one

fibre/plan.md
rach-id: Just wording issue 😆 no long-term direction 😆 
```suggestion
**Answer**: See DD2 for the decision. It's partially possible to do given the details in DD2.
```

rach-id:
> are there any other alternatives to solving the problem of allowing multiple PFFs in the same block?

IIUC, there can be multiple PFFs in the same block. The problem is that you can game it to only pay for the first, while the others will fail but still benefit from submitting data to fibre.

Another alternative could be a different version of the cache: query the account balance each block and decrement it locally and query it every block.

Another might be the fibre server queries the blocks and checks for included PFFs. If after sometime a promise is not included, they stop serving its corresponding data, instead of serving it for 24hrs.

fibre/plan.md
Wondertan: The long term direction is making PP validation stateless, i.e not requiring any IO

fibre/plan.md
Wondertan: It's likely to be a bottleneck given the Fibremaxing and synthetic storage benchmarks. We removed all the disk IO, which consists of the state machine and the actual blob storage.  The synthetic storage benchmarks for blob storage showed good performance, while state machine performance is in a grey spot. However, by the nature of LSM tree DBs, heavy reads are not performant, and considering we use the old Cosmos SDK Iavl tree implementation (please confirm whether we migrated), which is known for weak performance, it is valid to presume state is highly likely to be a bottleneck. Not to say we shouldn't confirm this empirically.

fibre/plan.md
Wondertan, line 16: This is an implementation detail. Let's not decide where the cache would live just yet. It is indeed very likely to be in fibre pkg, but I am not sure about the field on Server.

fibre/plan.md
```diff
@@ -0,0 +1,108 @@
...
+
+---
+
+## Design Decisions
+
+### DD1: Where does the cache live?
+**Answer**: In the fibre server (`fibre/` package), as a field on `Server`. The state machine remains the source of truth; the cache is a local optimization.
+
+### DD2: Do we still call `ValidatePaymentPromise` gRPC?
+**Answer**: Yes. The gRPC call handles non-balance stateful checks (timestamp freshness, expiration, height window, not-already-processed) that require chain state. The cache **only** replaces the balance sufficiency check. The gRPC call happens first; the local balance deduction happens after it succeeds.
```
Wondertan, line 19: Oh, wait, I assumed that with this, we would not call the RPC at all. Considering we still have to, I don't think this cache should live in fibre pkg. Logically, this whole problem is a state-machine problem, and by putting it in a fibre server, we expose state-machine details to the fibre server, which abstracts away validation, i.e. leaked abstraction

fibre/plan.md
Wondertan: However, I am talking here about disk IO being a bottleneck, not the actual GRPC call.

fibre/plan.md
```diff
@@ -0,0 +1,108 @@
...
+16. Invalid validator sign bytes
+17. Validator signature validation fails (can't get historical validator set, invalid signatures, or < 2/3 voting power)
+
+**Additional rejection for `MsgPaymentPromiseTimeout`** (checks 9-11 are skipped):
+18. Not yet timed out: `blockTime < expirationTime` (the inverse of check 9)
+
+**Answer**: So, if we implement the local validation of the payment promises. We can be sure that any payment promise that's submitted onchain and has enough signatures is valid. And we don't need to worry about this case. For the expired promises, that's handled in another issue.
+
+### DD8: Thread-safety approach?
+**Answer**: Single `sync.Mutex` on the cache. On cache hit: lock, deduct, unlock. On cache miss: lock to check, unlock, call gRPC (no lock held), lock to populate and deduct, unlock. Duplicate fetches on concurrent misses are harmless (both get the same block's state).
```
Wondertan, line 89: I think `sync.Map` would be a better choice for this type of cache, given the disjoint reads and writes and the fact that the number of entries only grows. This is a hot-path too, meaning there definitely will be an impact from reduced contention for concurrent blob submissions from multiple clients

fibre/plan.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Design Decisions
+
+### DD1: Where does the cache live?
+**Answer**: In the fibre server (`fibre/` package), as a field on `Server`. The state machine remains the source of truth; the cache is a local optimization.
+
+### DD2: Do we still call `ValidatePaymentPromise` gRPC?
+**Answer**: Yes. The gRPC call handles non-balance stateful checks (timestamp freshness, expiration, height window, not-already-processed) that require chain state. The cache **only** replaces the balance sufficiency check. The gRPC call happens first; the local balance deduction happens after it succeeds.
+
+**Proposal**: We could make many validations local. For example, the timestamp could be checked against the block time. We could throw events on params/governance changes and listen for those events so that when they happen, we update immediately. The goal would be to have as much of `ValidatePaymentPromise` logic running locally as possible, reducing reliance on the gRPC round-trip for each promise. This could be a follow-up.
```
Wondertan, line 21: Moving all the validation to the server is something I directionally agree with, but if the goal is to address RPC call overhead, I think it's an optimization that adds minimal value at significant complexity and time spent.  The end state is stateless verification that @nashqueue has talked about multiple times (there is an issue about it), and generally, I would be cautious about spending time on interim optimizations. 

In this case, we would need to copy a subset of the state into the fibre server to be managed by state(params/gov) events and new block events, essentially reimplementing the whole state machine within fibre, especially with point 5, which overall is lots of work and complexity to only reduce GRPC call overhead. If we find this is a real bottleneck, we could explore IPC or alternatives to GRPC before committing to moving state to the fibre server, because, again, that is lots of work that will be replaced with stateless verification anyway.

rach-id: I'm not worrying about the RPC overhead. I made this proposal to have all the validation in the same place, or at least most of it. Having the validation scattered in multiple places makes it harder to know what's happening.

Wondertan: In general, having everything in one place is better than having it scattered across two places(not multiple, btw, only two). But in this particular case, it's not gonna make things simpler as described above.

fibre/plan.md
rach-id: The problem I see with moving the cache to be in the state machine is adding a component that's not committed to + if the node's gRPC is used by multiple nodes, then it's hard to keep track of who signed which. 

Also, I see this is a responsability of the server to handle and not the state machine. The state machine's responsibility is processing the payment transactions and it's not aware of what happens between blocks. If the fibre server/protocol doesn't want to be used for free, then they should self-protect by keeping track of its signatures.

This is a state machine limitation and either the server or the state machine will have to deal with the smell.

fibre/plan.md
rach-id: If we have a plan for stateless validation, then that's way better and we should wait for it

fibre/plan.md
rootulp: That sounds like the problem we're trying to solve here. Do we need changes in the state machine to ensure that can't happen?

fibre/plan.md
Wondertan: > The problem I see with moving the cache to be in the state machine is adding a component that's not committed to

Wdym by adding a component that's not committed to?


> if the node's gRPC is used by multiple nodes, then it's hard to keep track of who signed which.

Who is who and what and why do the sign? 


BTW, I don't mean this cache would be literally in the state tree, just a subcomponent somewhere next to the keeper.

Wondertan: > This is a state machine limitation and either the server or the state machine will have to deal with the smell.


If this a state machine limitation, it means it should be fixed in the state machine and not hacked around by the dependants.

Imagine if a single core validator run multiple fibre servers (which is a very plausible scaling strategy we discussed), then every fibre server would have a duplicate cache instance that performs logically the same function for no good reason.

fibre/plan.md
rach-id: The problem happens between blocks. So as long as we allow multiple fibre blobs to be submitted between blocks, I don't think there is anything in the state machine that can help with this

fibre/plan.md
rach-id: > Imagine if a single core validator run multiple fibre servers 

If this is something we want, then yes it makes sense to have it somewhere in the validator instead of the fibre server. Not necessarily next to the keepers because that's state machine specific, but it could live somewhere separate. Probably next to the privval gRPC endpoint or something

Wondertan: > If this is something we want, then yes it makes sense to have it somewhere in the validator instead of the fibre server. 

The example I gave is an exercise/test one can do to see if the responsibilities/concerned were separated correctly. It is an exercise you can reapply for other architectural decisions as well BTW. 

It does not mean we have to decide we run multiple fibre servers per validator before putting cache next to the state. It means putting this to the fibre server is not sound.

evan-forbes: agree that the goal here is to make promise verification require the absolute minimum number of queries to the state machine (1 query to the statemachine per account per 23 hours)

we also don't need to include on the fibre server

fibre/plan.md
evan-forbes: > The end state is stateless verification

totally agree however I don't see how this can occur w/o boiling down to a similar solution here w/o zk proofs or state inclusion proofs

https://github.com/celestiaorg/fibre-da-spec/pull/16#discussion_r2410761744

it also doesn't address the problem of submitting more than one message per block

https://github.com/celestiaorg/fibre-da-spec/pull/16#discussion_r2410882972

fibre/plan.md
evan-forbes: another piece of important context @Wondertan is that we already need some second copy of the state to support submitting more than one promise per account per celestia block (realistically its once every 3 celestia blocks when accounting for the time to include the PFF onchain)

typically this is done via copying the entire state and applying using the normal state machine (slow and overly heavy)

the proposal here is to greatly simplify that second state machine

fibre/plan.md
```diff
@@ -0,0 +1,108 @@
...
+5. **Not already processed**: `IsPaymentPromiseProcessed()`. Needs processed payments KV store. Harder — need to track which promises have been included on-chain. Could read blocks or listen for `EventPayForFibre`/`EventPaymentPromiseTimeout` events.
+
+6. **Escrow account exists**. Needs escrow accounts KV store. Could be cached, seeded on first query, invalidated on deposits/withdrawals.
+
+7. **Sufficient available balance**: `AvailableBalance >= cost`. Needs escrow account balance. This is exactly what the escrow cache solves.
+
+Checks 1-4 only need `blockTime`, `blockHeight`, and `params` — all refreshable once per block. Check 5 is the hardest to make local. Checks 6-7 are what the escrow cache addresses. The proposal is to query this information once per block instead of once per-payment promise.
+
+### DD3: How does the cache get seeded with the on-chain balance?
+**Answer**: On cache miss, after the `ValidatePaymentPromise` gRPC call succeeds, make a separate `EscrowAccount` gRPC query to get the current `AvailableBalance`. The `EscrowAccount` request already exists and returns the balance.
```
evan-forbes, line 42: 👍

fibre/plan.md
evan-forbes, line 7: the root of the problem that we're trying to solve is how to enable users to submit more than one promise at a time.

this requires some form of state. If this is true, then how do we implement a state machine in a minimal, easy way.

Without the optimization that is described here, we would need basically maintain an entire mempool just for promises

this was done here actually! https://github.com/celestiaorg/celestia-app-fibre/pull/40

even then I tried to optimize it by avoiding the mempool, but it requires executing the entire promise state each celestia block (just like a mempool)!

fibre/plan.md
```diff
@@ -0,0 +1,108 @@
...
+8. Timestamp too old: `creationTime <= blockTime - WithdrawalDelay`
+9. Promise expired: `blockTime >= creationTime + PaymentPromiseTimeout`
+10. Height too far in past: `currentHeight - promiseHeight > PaymentPromiseHeightWindow`
+11. Height too far in future: `promiseHeight > currentHeight + 1`
+12. Already processed (promise hash in processed payments store)
+13. Escrow account not found
+14. Insufficient balance (`Balance < paymentAmount`)
+15. Insufficient available balance (`AvailableBalance < paymentAmount`)
+16. Invalid validator sign bytes
+17. Validator signature validation fails (can't get historical validator set, invalid signatures, or < 2/3 voting power)
```
evan-forbes, line 81: I think in a future document that is focused on the cache, we can literally just focus on strictly the payment verification, which you outlined above as being super super simple

if we want to expand this to being a holistic verification then I could see that making sense as well

check cache
if (no cache or funds are 0) && !CheckedThisHeight:
  - check state machine like normal

the && !CheckedThisHeight is optional to implement, but figured I'd include it here

fibre/plan.md
```diff
@@ -0,0 +1,108 @@
...
+**Answer**: So, if we implement the local validation of the payment promises. We can be sure that any payment promise that's submitted onchain and has enough signatures is valid. And we don't need to worry about this case. For the expired promises, that's handled in another issue.
+
+### DD8: Thread-safety approach?
+**Answer**: Single `sync.Mutex` on the cache. On cache hit: lock, deduct, unlock. On cache miss: lock to check, unlock, call gRPC (no lock held), lock to populate and deduct, unlock. Duplicate fetches on concurrent misses are harmless (both get the same block's state).
+
+### DD9: Should we limit one promise per account per height?
+**Answer**: No. This hurts UX for power users submitting multiple blobs. The balance cache is sufficient to prevent over-commitment without artificial rate limits.
+
+### DD10: Should we require dramatically higher escrow minimums?
+**Answer**: No. Same UX concern. The cache solves the problem without changing economic parameters.
```
evan-forbes, line 95: I'm not fully sure this actually solves the problem anyway

since they can always withdraw, which reduces the escrow account amount. therefore we'd need to do the exact same thing of cacheing the amount and decrementing it to get any beneft

evan-forbes (commented):
generally this design document has the info we want at a high level! however, imo not in the format we want (spec or more detailed implementation plan). 

Therefore, I don't think we should merge this document as is, but instead refactor the contents.

rach-id:
> however, imo not in the format we want (spec or more detailed implementation plan).
> 
> Therefore, I don't think we should merge this document as is, but instead refactor the contents.

100%, I only made this ready for review so that people review it. Even I wouldn't approve this to be merged 😆 But it puts us in the same picture, then will open an actual implementation plan along with the implementation

rach-id:
CLosing this as I'll open a separate PR with an implementation plan

fibre/plan.md
Wondertan: I agree that stateless requires more input from @nashqueue at this point. But I dont think we should be rushing with optimizing the validation either yet and focus on the cache/double spending first

PR #233, opened 2026-03-05
fix(fibre/grpc): sync.Once -> sync.Mutex to fix data race in ClientCache

Wondertan:
Close() read entry.clientCloser without holding the entry lock, racing with GetClient's write inside sync.Once.Do. Replace Once with Mutex so both methods synchronize on the same lock.

This was a flake on one of the recent PRs


rootulp (approved):

Wondertan:
This was merged without green CI, fyi

